PR #13, opened 2023-08-14
feat: MinGW compatible build.

tupoy-ya:
feat: Added GitHub CI.

Requires MinGW 13.1.0 or later.

Didn't test if it works in game, only loaded it in main menu once and the GUI didn't open.

##### TODO:
- [x] Fix `rage::scrNativeHash` to be unsigned in GTAV-Classes.
- [ ] Update `ghcr.io/yimmenu/yimmenu` to use latest MinGW.
- [x] Test if MinGW build actually works.

~~- [ ] Fix AsyncLogger not defining `USE_FMT`.~~
EDIT:
GCC 13+ Has `std::format` and the base is impossible to build with lower versions anyway.


maybegreat48:
We don't use the legacy branch anymore, and I've manually integrated most of your fixes into the codebase. Feel free open a new PR if you want to change some more stuff

PR #25, opened 2025-03-25
feat(BytePatch): add some bypass and ptr

AAA-ALR:


src/types/script/ScriptEvent.cpp
```diff
@@ -3,5 +3,6 @@
 
 void SCRIPT_EVENT::SendImpl(int size)
 {
-	SCRIPT::_SEND_TU_SCRIPT_EVENT_NEW(1, this, size / sizeof(std::uint64_t), PlayerBits, EventIndex);
+	YimMenu::Pointers.TriggerScriptEvent(1, (int64_t*)this, size / sizeof(std::uint64_t), PlayerBits);
```
maybegreat48: Is there a reason we're using this over the native?

AAA-ALR: I don't know either, I just found it and used this because on legacy yimmenu I remember using pointers instead of native

maybegreat48: As far as I can see they're doing the same thing, aside from the native trying to reset the header with the params you give it

AAA-ALR: Yes, I restored it now

maybegreat48:
Thanks! And I like the idea of storing byte patches in the pointers struct

PR #27, opened 2025-03-25
feat: improve vehicle spawner

danihamm:
Various improvements to the vehicle spawner:
  - Implement Vehicle::SetPlateText
  - Add license plate customization to vehicle spawner
  - Move vehicle upgrade/mod logic to Vehicle class (Vehicle::Upgrade)
  - Vehicle Spawner displays error notifications if model name is not provided or is invalid
  
  
image

### DIFF
--- a/src/game/gta/Vehicle.hpp
+++ b/src/game/gta/Vehicle.hpp
@@ -18,5 +18,7 @@ namespace YimMenu
 
 		// speed
 		float GetMaxSpeed();
+
+		void Upgrade();
 	};
 }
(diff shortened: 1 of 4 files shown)


theofficialhackclean:
Take the vehicle spawner and the vehicle options and put it in its own submenu and also find the vehicle so you use the class to list all the cars so you can spawn them using that, so you don't have to type in the name of the car also do the same for the weapons put them also in their own sub menu

src/game/gta/Vehicle.hpp
```diff
@@ -18,5 +18,8 @@ namespace YimMenu
 
 		// speed
 		float GetMaxSpeed();
+
+		void Upgrade();
+		void SetPlateText(const std::string text);
```
maybegreat48, line 23: I recommend using `std::string_view` instead of const strings to prevent unnecessary conversions when passing C-strings to the function. Just a note for next time

PR #32, opened 2025-03-25
Moved the code to their own submenu to make it cleaner and added player joining and leaving notifications

theofficialhackclean:


maybegreat48:
I'm not sure we need this yet. Maybe if there are like tens of options in each submenu but it's fairly clean now

theofficialhackclean:
@maybegreat48 
Yes, but makes it a lot easier because if you spawn in a car, you're in world then you have to go to self to do the car options and if you were if you use mine, it would just be one submenu and then if use the car spawner then all you would have to do is just switch tab instead of having to go out of the hole submenu

maybegreat48:
> if you spawn in a car, you're in world then you have to go to self to do the car options

All the car options we have so far are just toggles that you have to enable once and they'll apply across all vehicles

> all you would have to do is just switch tab instead of having to go out of the hole submenu

Again, I see your point, but by the time we have a remote LSC menu, we'll probably have a dedicated dropdown vehicle spawner in the Vehicle tab

theofficialhackclean:
@maybegreat48 
if you guys are going to added it later on why not just do now where you're adding everything so it would be easier and also it really doesn't make any sense to have it in self because it really not a self option

maybegreat48:
A lot has changed and this PR is stale now. If you really want to refactor the UI, feel free to open a new one

PR #33, opened 2025-03-26
feat(Self): enter last vehicle, infinite parachutes

danihamm:
Added 2 commands to Self:

- Infinite Parachutes - Player will always have a parachute and a reserve parachute.
- Enter last vehicle - Enters any free seat in the last vehicle the player was sat in.


G0KU0:
How can I create the dll because I can't do it?

theofficialhackclean:
please push my comment first then this one so we can move vehicle options to new submenu to be cleaner

theofficialhackclean:
So, everything is in one spot

PR #35, opened 2025-03-27
feat(Remove LSC Restrictions)

oakboat:
Removes Los Santos Customs vehicle modification restrictions

### DIFF
--- a/src/game/features/system/LSCCustomsBypass.cpp
+++ b/src/game/features/system/LSCCustomsBypass.cpp
@@ -40,5 +40,5 @@ namespace YimMenu::Features
 		}
 	};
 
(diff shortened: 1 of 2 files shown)


src/game/features/system/LSCCustomsBypass.cpp
```diff
@@ -0,0 +1,44 @@
...
+			}
+
+			if (m_BlockMenuOptionPatch)
+			{
+				m_BlockMenuOptionPatch->Disable();
+			}
+		}
+	};
+
+	static LSCCustomsBypass _LSCCustomsBypass{"lsccustomsbypass", "Remove LSC Restrictions", "Removes Los Santos Customs vehicle modification restrictions", true};
```
maybegreat48: I don't think we should enable this by default

oakboat: I think so and i fixed it

PR #64, opened 2025-03-31
Add Tunables Support

ShinyWasabi:


InfernoTsugikuni (commented):
I haven't really read the YimMenuV2 source code, but in my opinion modifying `IDLEKICK_KICK` seems pretty risky. Have you tested it and if so, is your account safe?

yigitboi07:
when this is going to be approved?

ShinyWasabi:
> I haven't really read the YimMenuV2 source code, but in my opinion modifying `IDLEKICK_KICK` seems pretty risky. Have you tested it and if so, is your account safe?

It is not risky at all. There is currently no detection mechanism for tracking the modification of globals (there used to be one in the past, though).

CSY0N:
Tunables and Lua support would be cool

InfernoTsugikuni:
> > I haven't really read the YimMenuV2 source code, but in my opinion modifying `IDLEKICK_KICK` seems pretty risky. Have you tested it and if so, is your account safe?
> 
> It is not risky at all. There is currently no detection mechanism for tracking the modification of globals (there used to be one in the past, though).

Thanks for letting me know, it's pretty dumb to leave globals unprotected but in that case i think that the owner should just merge this.

InfernoTsugikuni:
> when this is going to be approved?

No idea, probably when the owner notices this and checks if it's safe.

maybegreat48:
> It is not risky at all. There is currently no detection mechanism for tracking the modification of globals (there used to be one in the past, though).

It still exists but it only detects writes done outside script threads, and I'm not very sure if it actually reports it anywhere after all the AC changes

src/game/backend/Tunables.hpp
```diff
@@ -0,0 +1,81 @@
...
+			return GetInstance().m_CurrentJunkVal;
+		}
+
+		static int IncrementJunkVal()
+		{
+			return GetInstance().m_CurrentJunkVal++;
+		}
+
+		template <typename T>
+		static std::enable_if_t<std::is_pointer_v<T>, T> GetTunable(joaat_t hash)
```
maybegreat48: Can we make this return a `ScriptGlobal` instead? Then you can store it once instead of calling this every time

src/game/backend/Tunables.cpp
```diff
@@ -0,0 +1,199 @@
...
+					std::memcpy(ScriptGlobal(TUNABLE_BASE_ADDRESS).As<PVOID>(), m_TunablesBackup.get(), m_NumTunables * 8);
+
+					if (m_Tunables.size() == 0)
+					{
+						LOG(FATAL) << "Failed to cache tunables.";
+						return;
+					}
+
+					m_ScriptStarted = false;
+					m_Initialized   = true;
```
maybegreat48, line 140: We should clear `m_JunkValues` after the cache process to save some memory

src/game/backend/Tunables.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+			if (m_Initialized || m_Loading)
+				return;
+
+			if (!m_ScriptStarted && SCRIPT::GET_NUMBER_OF_THREADS_RUNNING_THE_SCRIPT_WITH_THIS_HASH("tuneables_processing"_J) > 0)
+				continue;
+
+			if (!m_ScriptStarted)
+			{
+				// Wait until main_persistent is started, executing our code in startup causes the game to crash for some reason
```
maybegreat48: It's because the MP global blocks haven't loaded yet. You can force-load them like you did `tunables_registration` if you want

src/game/features/self/NoIdleKick.cpp
```diff
@@ -0,0 +1,38 @@
...
+		using LoopedCommand::LoopedCommand;
+
+		const std::vector<joaat_t> m_TunableHashes = {"IDLEKICK_WARNING1"_J, "IDLEKICK_WARNING2"_J, "IDLEKICK_WARNING3"_J, "IDLEKICK_KICK"_J, "ConstrainedKick_Warning1"_J, "ConstrainedKick_Warning2"_J, "ConstrainedKick_Warning3"_J, "ConstrainedKick_Kick"_J};
+		const std::vector<int> m_DefaultValues = {120000, 300000, 600000, 900000, 30000, 60000, 90000, 120000};
+
+		virtual void OnTick() override
+		{
+			for (auto tunable : m_TunableHashes)
+			{
+				if (auto tunablePtr = Tunables::GetTunable<PINT>(tunable))
```
maybegreat48: Fairly pedantic but `int*` is preferred over PINT

src/game/backend/Tunables.cpp
ShinyWasabi: Oh, I see. I think we should just wait for startup to load them for us, instead of requesting bunch of registration scripts. There is almost no time difference anyway.

src/game/backend/Tunables.hpp
```diff
@@ -0,0 +1,90 @@
...
+		}
+
+		static ScriptGlobal GetTunable(joaat_t hash)
+		{
+			if (auto it = GetInstance().m_Tunables.find(hash); it != GetInstance().m_Tunables.end())
+			{
+				return ScriptGlobal(it->second);
+			}
+
+			return ScriptGlobal(TUNABLE_BASE_ADDRESS); // this is not good
```
maybegreat48, line 87: We should definitely do something about this later

PR #71, opened 2025-03-31
feat(ui): updated colors to be more clear

L7NEG:
Screenshot 2025-04-01 005256


CSY0N:
i like the now color much more fr

PR #91, opened 2025-04-03
SkipCutscene Update

CSY0N:
Added SkipCutscene what lots of ppl requested findable in the hotkeys section!


src/game/features/self/SkipCutscene.cpp
```diff
@@ -0,0 +1,18 @@
...
+	{
+		using Command::Command;
+
+		virtual void OnCall() override
+		{
+			CUTSCENE::STOP_CUTSCENE_IMMEDIATELY();
+		}
+	};
+
+	static skip_cutscene _skip_cutscene{"SkipCutscene", "SKIP_CUTSCENE", "SKIP_CUTSCENE_DESC"}; //static skip_cutscene _skip_cutscene{"SkipCutscene", "SKIP_CUTSCENE", "SKIP_CUTSCENE_DESC", 0};
```
maybegreat48: We don't use translations anymore so you have to put in the actual string instead of the label. Also please use `SkipCutscene` instead of `skip_cutscene` and remove the commented out line

src/game/features/self/SkipCutscene.cpp
```diff
@@ -0,0 +1,18 @@
+#include "core/commands/Command.hpp"
+#include "game/backend/Self.hpp"
+#include <game/gta/Natives.hpp>
```
maybegreat48: Local imports should be in quotes, like this
```cpp
#include "game/gta/Natives.hpp"
```

maybegreat48:
You should also add the skip cutscene command to the self menu, before suicide

CSY0N:
i have updated it

maybegreat48:
> i have updated it

Thanks! However you still have to
1. Add this command to the self menu (go to Self.cpp and duplicate the line adding the `suicide` command, then change one of the lines to `skipcutscene`)
2. Provide a new name and description that aren't translation labels from YimMenu (i.e. replace `SKIP_CUTSCENE` and `SKIP_CUTSCENE_DESC` with what you want the human-readable names to be)

L7NEG:
toolsGroup->AddItem(std::make_shared<CommandItem>("suicide"_J));
		toolsGroup->AddItem(std::make_shared<CommandItem>("SkipCutscene"_J)); just add this in frontend > submenus > self.cpp
		
		
@CSY0N

CSY0N:
done Bro @maybegreat48 i hope all is right and can i have your discord name  ? the writing with u DM? send me your name as PM

PR #111, opened 2025-04-06
Enums

CSY0N:
Added Enums


maybegreat48:
We already have most of these scattered around the codebase

PR #114, opened 2025-04-06
Add Money / RP

iR3SH:
Hello,

I've added 3 buttons allowing you to add:
- 10K / 100K RP
- 10M $

The code may not be in the best form. I'm new to C++, but it works. Use with caution. Don't use it multiple times in a row

### DIFF
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -1,28 +1,28 @@
 ﻿{
-    "configurations": [
-        {
-            "name": "x64-Release-MSVC",
-            "generator": "Ninja",
-            "configurationType": "RelWithDebInfo",
-            "buildRoot": "${projectDir}\\out\\build\\${name}",
-            "installRoot": "${projectDir}\\out\\install\\${name}",
-            "cmakeCommandArgs": "",
-            "buildCommandArgs": "",
-            "ctestCommandArgs": "",
-            "inheritEnvironments": [ "msvc_x64_x64" ],
-        },
-        {
-            "name": "x64-Release-Clang",
-            "generator": "Ninja",
-            "configurationType": "RelWithDebInfo",
-            "buildRoot": "${projectDir}\\out\\build\\${name}",
-            "installRoot": "${projectDir}\\out\\install\\${name}",
-            "cmakeCommandArgs": "",
-            "buildCommandArgs": "",
-            "ctestCommandArgs": "",
-            "inheritEnvironments": [ "clang_cl_x64_x64" ],
-            "enableClangTidyCodeAnalysis": true,
-            "clangTidyChecks": "clang-analyzer-*"
-        }
-    ]
+  "configurations": [
+    {
+      "name": "x64-Release-MSVC",
+      "generator": "Ninja",
+      "configurationType": "RelWithDebInfo",
+      "buildRoot": "${projectDir}\\out\\build\\${name}",
+      "installRoot": "${projectDir}\\out\\install\\${name}",
+      "cmakeCommandArgs": "",
+      "buildCommandArgs": "",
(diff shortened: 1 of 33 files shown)


ThrexeL:
Is this specifically for FSL users? I doubt it'll be pretty risky without it.
If not, how safe it is and what type of usage do you expect with "Don't use it multiple times in a row"

iR3SH:
I don't really know the limits, I just used the RP add-on several times in a row and once I logged back in I was reset to my old level. But if I use it 1 or 2 times then I change sessions, it's saved even if I turn off the game. I have FSL mode activated, I don't know if it works without it but I don't think it's a good idea to use it without it

maybegreat48:
Yeah, you can't add money this way unless you're using FSL, and there are some other issues with this code, such as truncation if your money is above $2.1 billion. I would just add this under general recovery options as an ImGuiItem and use something like `NETWORK_EARN_FROM_PICKUP` instead of playing with the money stat manually. Also make sure to only show this option to FSL users by guarding it with `AnticheatBypass::IsFSLEnabled` (you can look at the stat editor code to see what I mean by this)

Oh, one more thing. A general rule-of-thumb is to not export game logic from frontend files to the backend like you're doing now, and to instead create a dedicated file to hold any logic that needs to be shared by both backend and frontend code. Luckily for you, there's `game/gta/Stats.hpp` that does what you want. IMO, you don't need commands to add money or RP anyway. I'd just inline it in the recovery menu and call it a day

> "Don't use it multiple times in a row"

If you have FSL you can do whatever you want

iR3SH:
No problem, i will try to edit the code properly

iR3SH:
Tell me if it's better, by the way, when i try to use the Stats.cpp functions, the result don't work, so the money or the RP don't come to the player

CSY0N:
make it without FSL would be great

iR3SH:
I don't know if it's possible, for the moment I'm following the instructions to make the code correct, I'll finish it tomorrow. According to maybegreat48, it's not possible to add money without FSL mode

xAzurrr:
i tried adding it but getting this error ? anyone knows what im doing wrong ? 
image

edit) nvm i got it to work had to clone the rep again but idk why ?, anyways if there is someone who can help me with questions i have, can please reply thanks

GringoDevs:
I couldn't when I compile it becomes unknown

xZueS19:
> I couldn't when I compile it becomes unknown



> i tried adding it but getting this error ? anyone knows what im doing wrong ? image
> 
> edit) nvm i got it to work had to clone the rep again but idk why ?, anyways if there is someone who can help me with questions i have, can please reply thanks

are u guys using fsl ?

iR3SH:
> Yeah, you can't add money this way unless you're using FSL, and there are some other issues with this code, such as truncation if your money is above $2.1 billion. I would just add this under general recovery options as an ImGuiItem and use something like `NETWORK_EARN_FROM_PICKUP` instead of playing with the money stat manually. Also make sure to only show this option to FSL users by guarding it with `AnticheatBypass::IsFSLEnabled` (you can look at the stat editor code to see what I mean by this)
> 
> Oh, one more thing. A general rule-of-thumb is to not export game logic from frontend files to the backend like you're doing now, and to instead create a dedicated file to hold any logic that needs to be shared by both backend and frontend code. Luckily for you, there's `game/gta/Stats.hpp` that does what you want. IMO, you don't need commands to add money or RP anyway. I'd just inline it in the recovery menu and call it a day
> 
> > "Don't use it multiple times in a row"
> 
> If you have FSL you can do whatever you want

If i create a function in the Player class it's better no ? Then a group in the Recovery menu with Items ?

PR #117, opened 2025-04-07
feat(Various): 3 New Features

xesdoog:
- Self: Stand On Vehicles: This was requested a long time ago in Legacy YimMenu but was never added.
- Self: Disable Action Mode: Disables the player's forced weird stance during combat.
- Vehicle: Allow Hats Inside Vehicles: Prevents the player's headgear from being removed when they enter a vehicle.


CSY0N:
looks interesting but we will see what say @maybegreat48

xesdoog:
> looks interesting but we will see what say @maybegreat48

honest-work-meme-c7034f8bd7b11467e1bfbe14b87a5f6a14a5274b

PR #128, opened 2025-04-09
Fix read packed int not working

fangshun2004:
1. Fix read packed int not working
2. Remove the scope setting packed int (because packed int misoperation will damage the account and there is no need to modify it in bulk)
3. A configuration file that does not generate imgui.ini in the root directory of the game

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,23 +12,10 @@ include(cmake/minhook.cmake)
 
 # source
 set(SRC_DIR "${PROJECT_SOURCE_DIR}/src")
-
-function(GroupSourcesAndCollect curdir outputVar)
-    file(GLOB children RELATIVE ${CMAKE_SOURCE_DIR}/${curdir} ${CMAKE_SOURCE_DIR}/${curdir}/*)
-    foreach(child ${children})
-        if(IS_DIRECTORY ${CMAKE_SOURCE_DIR}/${curdir}/${child})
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -12,10 +12,23 @@ include(cmake/minhook.cmake)
 
 # source
 set(SRC_DIR "${PROJECT_SOURCE_DIR}/src")
-file(GLOB_RECURSE SRC_FILES
-    "${SRC_DIR}/**.hpp"
-    "${SRC_DIR}/**.cpp"    
-)
+
+function(GroupSourcesAndCollect curdir outputVar)
```
maybegreat48: I'm using Visual Studio and I don't need this to open the project and see the folder view. Just open the project dir as a folder instead of running CMake to generate a solution

src/game/frontend/Menu.cpp
```diff
@@ -111,6 +111,8 @@ namespace YimMenu
 	void Menu::SetupFonts()
 	{
 		auto& IO = ImGui::GetIO();
+		IO.IniFilename = NULL;
```
maybegreat48: You can make it store that in the project dir

fangshun2004: Is it necessary to keep the imgui configuration file in this project? (Save window size information)

maybegreat48: Yeah, I know some people wanted it on the Matrix server

PR #136, opened 2025-04-09
Fix Teleport to Waypoint

Cleanwn:
Prevents tp from continuing when there are no waypoint

### DIFF
--- a/src/game/features/teleport/TpToWaypoint.cpp
+++ b/src/game/features/teleport/TpToWaypoint.cpp
@@ -50,12 +50,12 @@ namespace YimMenu::Features
 
 		virtual void OnCall() override
 		{
-			if (!HUD::IS_WAYPOINT_ACTIVE())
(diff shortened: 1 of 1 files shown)


rocascent:
before i can tp out from places where you couldn't set a waypoint, but now it doesn't work😂

Cleanwn:
> before i can tp out from places where you couldn't set a waypoint, but now it doesn't work😂



> before i can tp out from places where you couldn't set a waypoint, but now it doesn't work😂

I understand what you mean. The current logic is to perform transmission only when there is a navigation point. As for your problem, as long as you can open the map through ESC, you can transfer (when you are indoors, you can choose to switch to the default map to mark the navigation point)

rocascent:
yeah, but in kosatka missions you cant set points. idk why, may be a bug?

--edit: i mean you must leave kosatka before you can set a point

PR #138, opened 2025-04-09
feat(network): Add Random Event Launcher

ShinyWasabi:


xZueS19:
yes finally , was looking for smth like this for maze bank shootout, great work @ShinyWasabi

aymenmerabta5:
Great work @ShinyWasabi to be like you what should i learn iam a fullstack web dev and thanks

PR #140, opened 2025-04-10
Feature/outfit editor

BegaMetaZone:
added outfit slot editing feature with the help of CoPilot , excuse my rusted skills I learned coding with pascal , UI integration might not be the best , again sry . I couldn't figure out why cmakelist didn't found the libs so I had to ..

### DIFF
--- a/src/game/frontend/submenus/OutfitEditor.cpp
+++ b/src/game/frontend/submenus/OutfitEditor.cpp
@@ -1,5 +1,4 @@
 #include "OutfitEditor.hpp"
-
 #include "core/frontend/manager/UIManager.hpp"
 #include "game/backend/Self.hpp"
 #include "game/frontend/items/Items.hpp"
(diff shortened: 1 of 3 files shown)


Psifter:
for cmake to find the lib* you just have to hit control s on the file

yigitboi07:
this should be a great feature if its implented correctly!

CSY0N:
underviewing waiting for response from @maybegreat48

CMakeLists.txt
```diff
@@ -14,7 +14,11 @@ include(cmake/minhook.cmake)
 set(SRC_DIR "${PROJECT_SOURCE_DIR}/src")
 file(GLOB_RECURSE SRC_FILES
     "${SRC_DIR}/**.hpp"
-    "${SRC_DIR}/**.cpp"    
+    "${SRC_DIR}/game/backend/OutfitManager.hpp"
```
maybegreat48: Please revert this

src/game/backend/OutfitManager.cpp
```diff
@@ -0,0 +1,56 @@
+#include "OutfitManager.hpp"
+
+#include "game/gta/Natives.hpp"
+
+namespace YimMenu
```
maybegreat48: You don't need an outfit manager, just put all the necessary logic under OutfitEditor and delete this file

src/game/frontend/GUI.cpp
```diff
@@ -96,4 +97,77 @@ namespace YimMenu
 			ScriptMgr::Yield();
 		}
 	}
+
+	void GUI::ShowOutfitAndPropEditor()
```
maybegreat48: You shouldn't add UI stuff to `GUI.cpp`. That's what the submenu files are for

src/game/frontend/GUI.cpp
```diff
@@ -96,4 +97,77 @@ namespace YimMenu
...
+	{
+		auto ped = PLAYER::PLAYER_PED_ID();
```
maybegreat48: Use `Self::GetPed` instead

src/game/frontend/GUI.cpp
```diff
@@ -96,4 +97,77 @@ namespace YimMenu
...
+		if (!ENTITY::DOES_ENTITY_EXIST(ped))
+		{
+			Notifications::Show("Error", "Player not found", NotificationType::Error);
```
maybegreat48: Use `ImGui::TextDisabled` to tell the user this

src/game/frontend/GUI.cpp
```diff
@@ -96,4 +97,77 @@ namespace YimMenu
...
+			return;
+		}
+
+		ImGui::Begin("Outfit and Prop Editor");
```
maybegreat48: You don't need a separate window

src/game/frontend/submenus/OutfitEditor.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "core/frontend/manager/Category.hpp"
+#include "game/frontend/GUI.hpp"
+#include "game/frontend/items/Items.hpp" 
+
+namespace YimMenu::Submenus
+{
+	OutfitEditor::OutfitEditor() :
+	    Submenu::Submenu("Outfit Editor")
+	{
+		// Create a shared_ptr for the category
```
maybegreat48: Make sure all UI stuff goes here and not in a separate window. The seemingly AI-generated comment above this line isn't very helpful

src/game/frontend/submenus/OutfitEditor.cpp
```diff
@@ -0,0 +1,23 @@
+#include "OutfitEditor.hpp"
+
+#include "core/frontend/manager/Category.hpp"
+#include "game/frontend/GUI.hpp"
+#include "game/frontend/items/Items.hpp" 
+
+namespace YimMenu::Submenus
+{
+	OutfitEditor::OutfitEditor() :
+	    Submenu::Submenu("Outfit Editor")
```
maybegreat48: Actually, on second thought, this shouldn't be a submenu at all. Add this as a new category under self (but still create a separate file)

src/game/frontend/GUI.cpp
```diff
@@ -96,4 +97,77 @@ namespace YimMenu
...
+		for (int i = 0; i < 12; ++i)
+		{
+			int drawable, texture;
+			OutfitManager::GetOutfitSlot(i, drawable, texture);
+
+			// Cache max values
+			const int maxDrawable = OutfitManager::GetMaxDrawable(i);
+			int maxTexture        = OutfitManager::GetMaxTexture(i, drawable);
+
+			const char* slotName = (i < 12) ? outfitSlotNames[i] : "Unknown";
```
maybegreat48: Unnecessary check since `i` can never exceed 11

src/game/backend/OutfitManager.cpp
maybegreat48: Or maybe add them to the `Ped` class

src/game/frontend/GUI.cpp
```diff
@@ -96,4 +97,77 @@ namespace YimMenu
...
+			OutfitManager::GetPropSlot(i, drawable, texture);
+
+			// Cache max values
+			const int maxDrawable = OutfitManager::GetMaxPropDrawable(i);
+			int maxTexture        = OutfitManager::GetMaxPropTexture(i, drawable);
+
+			const char* propName = propNames[i];
+
+			// Precompute labels
+			std::string drawableLabel = std::string(propName) + " Drawable##" + std::to_string(i);
```
maybegreat48: Slightly nitpicky but you can replace all your string concatenations with `std::format` and also get rid of the "cache" and "precompute" stuff because it still gets calculated every tick

BegaMetaZone: last one didn't have the cache and it wasn't snappy , it had a delay without caching but alright I'll try to implement these changes .

xAzurrr:
is there a way to implement a tattoo editor ?. i looked into but nothing seems to work,. also they dont overlap

BegaMetaZone:
I was wondering that myself, have 2 check

Deadlineem:
This is a great feature IF it is properly implemented.  What I mean by this is, YimMenu V1 had an outfit editor, it was a bit buggy and not labeled correctly, also, it was stuck in one position on the screen (So you could not see your character with the menu opened) resulting in having to show/hide the menu for each outfit item you changed.

I changed this in YimMenu V1 (Chronix) so the outfit editor is movable/resizable and fixed all the mismatched labeling as well as added tabs (Clothing/Accessories) for a better user experience.

As far as the labeling went, the original outfit editor was, to my own guess, built in story mode, which made the clothing item labels "correct" in the menu, but for MP, they were wrong.

https://github.com/Deadlineem/Chronix/blob/main/src/views/self/view_outfit_editor.cpp

BegaMetaZone:
Before moving it under self this had separate window and it was editable but now it's under self and follows top UI properties which is fine , also the labels can be edited but for me they are fine rn , I didn't want to force outfits so if game doesn't like ur selection ( invis body ) it will automatically fix it after couple of seconds , I added Outfit Preset Saving in my local branch plus tattoo editing but tattoo isn't working rn , does not draw . about the preset saving , since it wasn't persistent when I reloaded the game I didn't push it .

Deadlineem:
> Before moving it under self this had separate window and it was editable but now it's under self and follows top UI properties which is fine , also the labels can be edited but for me they are fine rn , I didn't want to force outfits so if game doesn't like ur selection ( invis body ) it will automatically fix it after couple of seconds , I added Outfit Preset Saving in my local branch plus tattoo editing but tattoo isn't working rn , does not draw . about the preset saving , since it wasn't persistent when I reloaded the game I didn't push it .

To make it persist you need to save the outfit files in .json like yimv1 does (saved_outfits folder) and then the persist outfit would apply on a loop from that ,json outfit.

I personally did not like having the editor in the main UI on YimV1 bc it was so bulky, but i think it'd be ok with this UI as it doesn't overlay over your character.

qwertyuiop11243:
> > Before moving it under self this had separate window and it was editable but now it's under self and follows top UI properties which is fine , also the labels can be edited but for me they are fine rn , I didn't want to force outfits so if game doesn't like ur selection ( invis body ) it will automatically fix it after couple of seconds , I added Outfit Preset Saving in my local branch plus tattoo editing but tattoo isn't working rn , does not draw . about the preset saving , since it wasn't persistent when I reloaded the game I didn't push it .
> 
> To make it persist you need to save the outfit files in .json like yimv1 does (saved_outfits folder) and then the persist outfit would apply on a loop from that ,json outfit.
> 
> I personally did not like having the editor in the main UI on YimV1 bc it was so bulky, but i think it'd be ok with this UI as it doesn't overlay over your character.

I mean invisible clothing isn't a big deal and I doubt anyone actually likes how it looks, I personally think the loop should be left out of it. No need to please the public lobby rats. The game also already has a saved outfits option, the source is already out there so I doubt it would be difficult to pull from that instead of JSON.

BegaMetaZone:
yeah exactly what I was going for , since game already has outfit saving system it's not actually needed and about the invis parts or exclusive mask hat combo or long sleeves , I don't want to force a loop of applying outfit .

CSY0N:
hm will this be added to the menu? 🤨

L7NEG:
maybe this pr will be enough ? https://github.com/BegaMetaZone/YimMenuV2/pull/1

src/game/frontend/submenus/OutfitEditor.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include "core/frontend/manager/Category.hpp"
+
+namespace YimMenu::Submenus
+{
+	class OutfitEditor
+	{
+	public:
+		static std::shared_ptr<YimMenu::Category> CreateCategory();
+
+		static void SetOutfitSlot(int slot, int drawable, int texture);
```
maybegreat48: These don't need to be in the header file

src/game/frontend/submenus/OutfitEditor.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+#include <memory>
+#include "core/frontend/manager/Category.hpp"
+
+namespace YimMenu::Submenus
+{
+	class OutfitEditor
+	{
+	public:
+		static std::shared_ptr<YimMenu::Category> CreateCategory();
```
maybegreat48: Should be `CreateOutfitsMenu` instead, and get rid of the OutfitEditor namespace

L7NEG: Sure I'm working on it

L7NEG:
Done ✅ https://github.com/BegaMetaZone/YimMenuV2/pull/2

xAzurrr:
did you manage to add tatto editor ? i gave up to add it

src/game/frontend/submenus/OutfitEditor.hpp
```diff
@@ -0,0 +1,16 @@
+#pragma once
+#include "core/frontend/items/Category.hpp"
```
xZueS19, line 2: this path should be #include "core/frontend/managerCategory.hpp"

GalisWCar:
> did you manage to add tatto editor ? i gave up to add it

I was able to draw tattoos but there were complications , first was the huge json list of tattoos which I couldn't convert to hardcoded or make a deserializer for it , second , R* was drunk that day and didn't make any remove single overlay native which makes it that you need a system where for removing a single tattoo you needed to have a array of applied tattoos , remove that single one from list and apply the list all over again , and there was another issue , I couldn't find where R* stores tattoo or how they do it  so tattoos don't save and If you apply a wrong tattoo your character gets broken , in short yes , but it's better be implemented by someone with more skills than me .

src/game/frontend/submenus/OutfitEditor.hpp
L7NEG: mb noted ✅️

L7NEG:
Thx @xZueS19 fixed ✅ https://github.com/BegaMetaZone/YimMenuV2/pull/3

CSY0N:
@L7NEG you should open your own Pull Request that it get merged. on the official YImMenuv2 github page

L7NEG:
> @L7NEG you should open your own Pull Request that it get merged. on the official YImMenuv2 github page

soon but it is not a problem to help other members too :D

src/game/frontend/submenus/OutfitEditor.hpp
```diff
@@ -0,0 +1,16 @@
+#pragma once
+#include "core/frontend/manager/Category.hpp"
+
+namespace YimMenu
+{
+    int GetMaxDrawable(int slot);
```
maybegreat48: These don't need to be exposed in the header file. Other than that LGTM

L7NEG: Done Thx for noticing it ✅ https://github.com/BegaMetaZone/YimMenuV2/pull/4

L7NEG:
ngl this https://github.com/YimMenu/YimMenuV2/pull/234 is cool feature but it wont just invalidate this Custom outfit editor 
like the pics here are little diff 

Open Wardrobe Preview:
Screenshot 2025-04-24 023854


and 

Outfit Editor Preview:
Screenshot 2025-04-24 023926

Sylvo6642:
why is this not added yet

PR #153, opened 2025-04-11
use `if` instead of `goto`

rocascent:
btw maybe we can use spdlog for logging and stacktrace. idk which is better.


xiaoxiao921:
ah yes, rewriting code just for the sake of rewriting, a classic

https://jorenar.com/blog/gotophobia-harmful

Yimura:
There's no need to throw `goto` statements away necessarily, there are times to use it and times to not use it.

PR #180, opened 2025-04-13
feat: Format Money

ShinyWasabi:


src/game/features/self/FormatMoney.cpp
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+
+	namespace Hooks
+	{
+		void Features::FormatInt(int64_t integer, char* str, size_t strSize, bool insertCommas)
+		{
+			auto retnAddr  = _ReturnAddress();
+			auto retnBytes = static_cast<unsigned char*>(retnAddr);
+			if (YimMenu::Features::_FormatMoney.GetState() && retnBytes[0] == 0x0F && retnBytes[1] == 0x28 && retnBytes[2] == 0x05)
```
maybegreat48: You should probably use a `CallHook` instead to hook a function call at its call site

PR #193, opened 2025-04-15
Add Random Event ESP & Stuff

ShinyWasabi:
- Added random event ESP.
- Fixed kill random event.
- Added force script host.


src/game/backend/ScriptPatches.cpp
```diff
@@ -266,4 +267,18 @@ namespace YimMenu
 		m_Patches.clear();
 		m_Datas.clear();
 	}
+
+	void ScriptPatches::RunScriptImpl()
+	{
+		std::vector<ScriptPatch> sendUpdateRECoordsTSECooldownPatches{};
+
+		if (sendUpdateRECoordsTSECooldownPatches.empty())
```
maybegreat48: Add these when you need them to work and enable them once, not whenever the menu starts up
```cpp
static bool run_once = [] {
     // run code here
}();
```

src/game/gta/Stopwatch.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+struct TIMER;
+
+namespace YimMenu::Stopwatch
```
maybegreat48: This should be in `types/script`, under the TIMER class

src/core/util/Math.hpp
```diff
@@ -18,4 +18,9 @@ namespace YimMenu::Math
 
 		return rage::fvector3{static_cast<float>(-sin(z) * num), static_cast<float>(cos(z) * num), static_cast<float>(sin(x))};
 	}
+
+	inline float DistanceBetweenVectors(rage::fvector3 a, rage::fvector3 b)
```
maybegreat48: There's a method in rage::vector3 that does this for you

PR #211, opened 2025-04-19
Cache Script Pointers

ShinyWasabi:
TO-DO:
- Cache script patches as well (DONE).
- Trigger a cache save if a new pointer is added (DONE).


src/types/script/globals/Globals.cpp
```diff
@@ -13,26 +13,29 @@
...
-DEFINE_GLOBAL_ACCESSOR(GSBD, 2648917);
-DEFINE_GLOBAL_ACCESSOR(GSBD_BlockB, 2650415);
-DEFINE_GLOBAL_ACCESSOR(GSBD_FM, 1835557);
-DEFINE_GLOBAL_ACCESSOR(GSBD_FM_Events, 1917230);
-DEFINE_GLOBAL_ACCESSOR(GSBD_Kicking, 1877529);
-DEFINE_GLOBAL_ACCESSOR(GSBD_PropertyInstances, 1937699);
-DEFINE_GLOBAL_ACCESSOR(g_AMC_playerBD, 2699624);
-DEFINE_GLOBAL_ACCESSOR(GSBD_RandomEvents, 1882524);
\ No newline at end of file
+DEFINE_GLOBAL_ACCESSOR(GlobalPlayerBD, "GlobalPlayerBD"_J);
```
maybegreat48: I don't think these should be pointers, since it's so easy to update them, and besides, the structs would have to be updated anyway

src/game/pointers/ScriptPointers.cpp
```diff
@@ -0,0 +1,182 @@
...
+			auto shop_controller       = Scripts::FindScriptProgram("shop_controller"_J);
+			auto fm_mission_controller = Scripts::FindScriptProgram("fm_mission_controller"_J);
+			auto AM_MP_VEHICLE_REWARD  = Scripts::FindScriptProgram("AM_MP_VEHICLE_REWARD"_J);
+			if (!freemode || !shop_controller || !fm_mission_controller || !AM_MP_VEHICLE_REWARD)
+				continue;
+
+			ScriptPointer scanner{};
+
+			scanner = ScriptPointer(freemode);
+			AddPointer("GlobalPlayerBD"_J, scanner.Scan("61 ? ? ? 49 ? ? 46 ? ? 42 ? 5D ? ? ? 56 ? ? 5D").Add(1).Rip().As<std::uint32_t>());
```
maybegreat48: What if you still have the patterns where you use them, and then you proceed to cache them there after you find one for the first time? Something you will notice is that the script isn't always loaded when you create these pointers (like most script patch inits), so you will have to create a symbolic `ScriptPointer` that tracks all the manipulations you need to make, then you store that and scan it when the script is ready

src/game/frontend/submenus/network/RandomEvents.cpp
```diff
@@ -110,6 +110,38 @@ namespace YimMenu::Submenus
...
 
+	static void KillActiveEvent()
+	{
+		if (auto eventThread = Scripts::FindScriptThread(randomEventScripts[(int)selectedEvent]))
+		{
+			if (auto NetComponent = reinterpret_cast<GtaThread*>(eventThread)->m_NetComponent)
+			{
+				if (auto eventProgram = Scripts::FindScriptProgram(randomEventScripts[(int)selectedEvent]))
+				{
+					// TO-DO: Cache these too?
```
maybegreat48, line 120: And then you'll be able to cache these automatically, and most of the script patches too (except the weird hack I had to use to scan the same pattern in four different locations)

src/game/gta/ScriptFunction.cpp
```diff
@@ -1,34 +1,87 @@
...
 namespace YimMenu
 {
-	ScriptFunction::ScriptFunction(const std::string& name, const joaat_t script, const SimplePattern& pattern, std::int32_t offset, bool rip) :
-		m_Name(name),
-		m_Script(script),
-		m_Pattern(pattern),
-		m_Offset(offset),
-		m_Rip(rip),
-		m_PC(0)
+	ScriptFunction::ScriptFunction(const joaat_t nameHash, const joaat_t script) :
```
maybegreat48: The PC should still always go in the ScriptFunction ctor, or a ScriptPointer after my changes are made

src/game/pointers/ScriptPointers.cpp
```diff
@@ -0,0 +1,182 @@
...
+		return loaded;
+	}
+
+	ScriptPointers::ScriptPointer::ScriptPointer(rage::scrProgram* program) :
+	    m_Program(program),
+	    m_Address(0)
+	{
+	}
+
+	ScriptPointers::ScriptPointer::ScriptPointer(rage::scrProgram* program, std::uint32_t address) :
```
maybegreat48: This should just have the program bytes, and not the whole program

src/game/pointers/ScriptPointers.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+		void RunScriptImpl();
+		void Save();
+		void Load();
+
+		CacheFile m_CacheFile;
+		std::unordered_map<joaat_t, std::uint32_t> m_ScriptPointers{};
+	};
+
+	using scrPtr = ScriptPointers::ScriptPointer;
```
maybegreat48: using ScriptPointer = ScriptPointers::Pointer

src/game/pointers/ScriptPointers.hpp
```diff
@@ -0,0 +1,85 @@
+#pragma once
```
maybegreat48: I'm not sure if this should be in `pointers`, since that's for game pointers, not script stuff. This should probably be in `backend`, or maybe `gta` if you are willing to split the pattern-scanning and caching functionality

src/game/pointers/ScriptPointers.cpp
ShinyWasabi: You mean I should perform the scanning process on target script load (i.e. in `InitNativeTables`)?
```cpp
ScanPointerOnLoad("shop_controller"_J, [](rage::scrProgram* program) {
	ScriptPointer ptr{program};
	AddPointer("SendToClouds"_J, ptr.Scan("2D 00 02 00 00 72 5D ? ? ? 72").As<std::uint32_t>());
});
```

src/types/script/globals/Globals.cpp
ShinyWasabi: I still think that we should at least cache the base addresses, there is no harm in it after all.

src/game/pointers/ScriptPointers.cpp
maybegreat48: Well that, and keep them where you use them, instead of in a single init func

PR #217, opened 2025-04-20
Changed control

Cleanwn:
When controlling character movement, character behavior will not be triggered


maybegreat48:
Can you describe what you're trying to do? I'm not sure what you mean by "character behavior"

Cleanwn:
> 你能描述一下你想做什么吗？我不确定你所说的“性格行为”是什么意思



> Can you describe what you're trying to do? I'm not sure what you mean by "character behavior"



> Can you describe what you're trying to do? I'm not sure what you mean by "character behavior"

1. When moving, the new code will keep the character in a static state, and the action triggered by the character when moving will not occur. 2. A code that uses spaces to accelerate has been added, there is a default speed and acceleration speed.

PR #218, opened 2025-04-20
Radio Station on Foot

CSY0N:
Added Mobile Radio


maybegreat48:
I guess I'll merge this, but be sure to take my advice from my previous PR and learn how the code works. For example, a quick glance at the file tree would tell you that you're supposed to use CamelCase for file names, not snake case. If you browse a few files under `features`, you can also see that you're not supposed to use translation labels as your command name. I would recommend just looking through the code first before trying to add any more features, and improving your English skills so we can communicate easier

PR #221, opened 2025-04-21
Anticheat bypass adapted for FSL v2

SourceModzZ:
Since there will now be FSL v2, and the .dll will no longer be called “version” but “WINMM”, here is an adjustment.


L7NEG:
interesting part is that fsl will run with BE on since it has a bypass for it or not its not yet confirmed ?
image

PR #223, opened 2025-04-21
Add Support for FSL v2 and API intigration

SourceModzZ:


src/game/backend/AnticheatBypass.hpp
```diff
@@ -15,6 +15,9 @@ namespace YimMenu
 
 		bool m_IsFSLLoaded = false;
 		bool m_BattlEyeRunning = false;
+		int m_FSLVersion       = -1;
```
maybegreat48: You should expose these through AnticheatBypass methods and replace existing uses of IsFSLEnabled to work with the new interfaces

SourceModzZ:
I think that's all right now.

maybegreat48:
OK let's wait for FSL to be approved and see if it works

Skorchekd:
this may be kind of unrelated, but what does fsl 2 really add?, from what ive read it only makes it so that you can use that fsl on both versions, no advancements made to mitigate battle-eye detection in online so we can play with other players? just a question i have, apologies if its unrelated.

SourceModzZ:
> this may be kind of unrelated, but what does fsl 2 really add?, from what ive read it only makes it so that you can use that fsl on both versions, no advancements made to mitigate battle-eye detection in online so we can play with other players? just a question i have, apologies if its unrelated.

FSL 2 primarily allows you to activate and deactivate local storage on the PC via config. It is also possible to deactivate and activate the battleye bypass via config. The developers have integrated an API which developers can call up with menus, e.g. fsl version (which is rather pointless), call up whether local saves & BattlEye bypass is activated or deactivated.

Skorchekd:
> > this may be kind of unrelated, but what does fsl 2 really add?, from what ive read it only makes it so that you can use that fsl on both versions, no advancements made to mitigate battle-eye detection in online so we can play with other players? just a question i have, apologies if its unrelated.
> 
> FSL 2 primarily allows you to activate and deactivate local storage on the PC via config. It is also possible to deactivate and activate the battleye bypass via config. The developers have integrated an API which developers can call up with menus, e.g. fsl version (which is rather pointless), call up whether local saves & BattlEye bypass is activated or deactivated.

so fsl 2 adds a bypass? or is the heartbeat still an issue? it seems theyve just added an api and variables. Apologies, just abit confused.

SourceModzZ:
FSL V2 has been approved. Unfortunately, it doesn't work on Enhanced, so I can't test the API.

boxlemons21:
> FSL V2 has been approved. Unfortunately, it doesn't work on Enhanced, so I can't test the API.

Bro thats crazy, rockstar patched way too quickly because UC mods r lazy fks

SourceModzZ:
> > FSL V2 has been approved. Unfortunately, it doesn't work on Enhanced, so I can't test the API.
> 
> Bro thats crazy, rockstar patched way too quickly because UC mods r lazy fks

Yes, that is definitely true.

SourceModzZ:
A small mistake had crept in after my last change. I have added BattlEyeStatusUpdatePatch again because it was missing.

xZueS19:
> > FSL V2 has been approved. Unfortunately, it doesn't work on Enhanced, so I can't test the API.
> 
> Bro thats crazy, rockstar patched way too quickly because UC mods r lazy fks

image
stop saying random shit

Skorchekd:
> > > FSL V2 has been approved. Unfortunately, it doesn't work on Enhanced, so I can't test the API.
> > 
> > 
> > Bro thats crazy, rockstar patched way too quickly because UC mods r lazy fks
> 
> image stop saying random shit

🤔 Did they add a bypass to stop kicks in new fsl

xZueS19:
> > > > FSL V2 has been approved. Unfortunately, it doesn't work on Enhanced, so I can't test the API.
> > > 
> > > 
> > > Bro thats crazy, rockstar patched way too quickly because UC mods r lazy fks
> > 
> > 
> > image stop saying random shit
> 
> 🤔 Did they add a bypass to stop kicks in new fsl

it will never gets added unless someone figure out how to deal with the battleye handshake

PR #257, opened 2025-04-30
clear damage and skip line in phone conversation commands.

rohitlalwani2812-personal:
change clear damage into toggle command.
add skip conversation command which skips current line of phone conversation upon activation.

### DIFF
--- a/src/game/features/self/ClearDamage.cpp
+++ b/src/game/features/self/ClearDamage.cpp
@@ -1,17 +1,14 @@
-#include "core/commands/LoopedCommand.hpp"
+#include "core/commands/Command.hpp"
 #include "game/backend/Self.hpp"
 
 namespace YimMenu::Features
 {
-	class ClearDamage : public LoopedCommand
+	class ClearDamage : Command
 	{
-		using LoopedCommand::LoopedCommand;
+		using Command::Command;
 
-		virtual void OnTick() override
+		virtual void OnCall() override
 		{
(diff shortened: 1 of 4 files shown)


src/game/features/self/ClearDamage.cpp
```diff
@@ -1,14 +1,17 @@
-#include "core/commands/Command.hpp"
+#include "core/commands/LoopedCommand.hpp"
 #include "game/backend/Self.hpp"
 
 namespace YimMenu::Features
 {
-	class ClearDamage : Command
+	class ClearDamage : public LoopedCommand
```
maybegreat48: This is just godmode

rohitlalwani2812-personal: Not godmode.  But clears visual blood on the player and keeps it that way.

maybegreat48: Never mind, you're right. I'd still keep the button and add a new looped command for this

rohitlalwani2812-personal: Wont that be redundant though?  Having two similar commands?

maybegreat48: One is something you can run once and the other is looped. We have the same thing for vehicle fix

rohitlalwani2812-personal: @maybegreat48 updated the branch.

PR #259, opened 2025-05-01
add hostkick

latiao-1337:
add hostkick


SourceModzZ:
I don't think it will work. The kick on this native has been patched for a long time.

latiao-1337:
> I don't think it will work. The kick on this native has been patched for a long time. 

No, it will work, but only when you are the host.

PR #263, opened 2025-05-01
object esp, enemy relationship fix, teleport to objective, delete camera

rohitlalwani2812-personal:


src/game/gta/Entity.cpp
```diff
@@ -421,22 +421,27 @@ namespace YimMenu
...
-			std::uint32_t weapon = "WEAPON_EXPLOSION"_J;
-
-			if (!ptr || !local)
-				return;
+		if (IsDead())
+			return;
 
-			Pointers.TriggerWeaponDamageEvent(local, ptr, &pos, 0, true, weapon, 9999.9f, 2, 0, (1<<4)|0x80000, 0, 0, 0, false, false, true, true, nullptr);
-		}
+		auto weaponHash  = Self::GetPed().GetCurrentWeapon();
```
maybegreat48: The damage event is much more reliable than sending fake bullets through the network

rohitlalwani2812-personal: Yes. But the event doesn’t register any kill in the stats or the mission.
Edit: will once double check with trigger weapon damage event method and see if it registers the kill as well.  If not, this bullet simulation works fine as well I noticed.  Doesn't kill every time due to peds being blocked by vehicle or other things, but every kill is registered by the game.

src/game/frontend/ESP.cpp
```diff
@@ -269,6 +280,39 @@ namespace YimMenu
...
+		else if (distance > 100.f && distance < 300.f)
+			color = Orange;
+		else if (distance > 300.f)
+			color = Red;
+		std::string unit        = (distance < 1000.0f) ? "m" : "km";
+		auto objectHash = object.GetModel();
+		std::string objectName = "0x" + std::to_string(objectHash);
+		if (m_objectNames.find(objectName) != m_objectNames.end())
+		{
+			objectName = m_objectNames[objectName] + " (" + objectName + ")";
```
maybegreat48: Why are you converting the hash into a string before trying to look it up?

src/game/frontend/ESP.cpp
```diff
@@ -269,6 +280,39 @@ namespace YimMenu
...
 
+	void ESP::DrawObject(Entity object, ImDrawList* drawList)
+	{
+		if (HUD::IS_PAUSE_MENU_ACTIVE() || NETWORK::NETWORK_IS_IN_MP_CUTSCENE())
+			return;
+
+		if (!object)
+			return;
+
+		if (!object.IsObject())
```
maybegreat48: I don't think you need this check

src/game/frontend/ESP.cpp
```diff
@@ -168,7 +179,7 @@ namespace YimMenu
 
 	void ESP::DrawPed(Ped ped, ImDrawList* drawList)
 	{
-		if (!ped.IsValid() || ped.IsPlayer() || ped == Self::GetPlayer().GetPed() || worldToScreen(ped.GetBonePosition(torsoBone)).x == 0 || (ped.IsDead() && !Features::_ESPDrawDeadPeds.GetState()))
+		if (!ped || !ped.IsValid() || ped.IsPlayer() || ped == Self::GetPlayer().GetPed() || worldToScreen(ped.GetBonePosition(torsoBone)).x == 0 || (ped.IsDead() && !Features::_ESPDrawDeadPeds.GetState()))
```
maybegreat48: `!ped` is equivalent to `!ped.IsValid()`

src/game/frontend/ESP.cpp
rohitlalwani2812-personal: I have found that peds are also part of the objects pool.  Thus, I added this check to only draw if it is actually an object.

src/game/frontend/ESP.cpp
rohitlalwani2812-personal: As I was already changing to string to have it as default text shown, I was using it in the map too.
But I agree that it is not ideal.  I will change so that I look up using hash itself.

src/game/frontend/ESP.cpp
rohitlalwani2812-personal: Ok. I will once remove and test this out.
Why I added this check -> At times when using Ped ESP, I noticed getting assert warnings in the console for GetBonePosition.  Upon adding this here and in other places, it works well now.

src/game/gta/Entity.hpp
```diff
@@ -90,6 +90,9 @@ namespace YimMenu
 		bool IsObject();
 		bool IsPlayer();
 		bool IsMissionEntity();
+		bool IsCCTV();
```
maybegreat48: These should be in `Object.hpp`. Also I'd call them `IsCamera` `IsCache` `IsSignalJammer`

src/game/gta/Entity.cpp
rohitlalwani2812-personal: Update: i am reverting this change.

src/game/frontend/ESP.cpp
rohitlalwani2812-personal: reverting this too.  as it is same.

src/game/frontend/ESP.cpp
rohitlalwani2812-personal: this is needed for above mentioned reason

src/game/frontend/ESP.cpp
rohitlalwani2812-personal: updating this in my code already. will push the latest branch today/tomorrow.

L7NEG:
simulated bullet can be just nammed bullet track like pubg mobile

rohitlalwani2812-personal (commented):
@maybegreat48 please let me know if any other changes are required here.

src/game/features/teleport/TpToObjective.cpp
```diff
@@ -0,0 +1,56 @@
...
+				if (GetBlipLocationOfType(location, sprite))
+				{
+					Self::GetPed().TeleportTo(location);
+					return;
+				}
+			}
+		}
+	};
+
+	class TpToArcade : public Command
```
maybegreat48: Why is this here? Also we don't really need this feature anyway since you can just set a waypoint to your arcade and TP to it

src/game/gta/Entity.cpp
```diff
@@ -421,6 +421,15 @@ namespace YimMenu
 	{
 		ENTITY_ASSERT_VALID();
 
+		std::unordered_set<long long> enemiesNotToKill = {
```
maybegreat48: This check should not be here

maybegreat48:
Other than that, LGTM

src/game/features/teleport/TpToObjective.cpp
rohitlalwani2812-personal: what you said is true but this is so that we can quick teleport to known locations.  i will be adding other locations in next branch/pr.  let me know if that's fine.

src/game/gta/Entity.cpp
rohitlalwani2812-personal: should it be in is enemy method then? 
or in kill all enemies command?

src/game/features/teleport/TpToObjective.cpp
maybegreat48: Yeah, but if you want to do that, you should add support to teleport to all active blips by iterating through them and displaying them in a menu

src/game/gta/Entity.cpp
maybegreat48: The check should be in the kill all enemies command yes. Also why do we need this anyway?

rohitlalwani2812-personal: this is so that in some missions when kill all enemies or kill all peds is used, the main enemy is not killed and the mission doesn't fail.  example is dr dre final mission.  one of setup in dr dre.

src/game/features/teleport/TpToObjective.cpp
rohitlalwani2812-personal: ok. i will remove this for now.  in other branch i will add it as dropdown like you said.

rohitlalwani2812-personal:
closing this due to conflicts and older code.  i will create new pr with latest changes along with feedback related ones.

TT-HK:
寻求会写CF/PUBG游戏外挂作者合作，你负责制作我负责售卖

PR #269, opened 2025-05-02
fixed cayo perico heist secondary loot causes transaction error

kurumimeow:
Reduced Cayo Perico Heist secondary loots value to avoid transaction error.


sweervin:
can I unload the menu after modify the secondary values?

PR #303, opened 2025-05-12
fix(Clang&MinGW): Fix Clang and MinGW builds on linux.

tupoy-ya:
This likely breaks building with clang for windows-gnu target, but it probably doesn't work anyway.

Updates `minhook` and `nlohmann::json` to work on latest cmake.
Adds clangd's `.cache` directory to `.gitignore`.

~**This was not tested in game!**~
Should work in game.


src/types/network/netEvent.hpp
```diff
@@ -28,6 +28,7 @@ namespace rage
 		std::uint32_t m_CxnId;                 // 0x0044
 		rage::netEvent* m_This;                // 0x0048
 		uint32_t m_PeerId;                     // 0x0050
+		char pad_0084[4];                      // 0x0054
```
maybegreat48, line 31: Wonder why GCC wants padding here. Might be a compiler bug

PR #326, opened 2025-05-19
Vehicle Stuff

ShinyWasabi:
- Added feature "Call Mechanic".
- Added safety check to prevent saving forbidden vehicles.
- Added vehicle spawner with search box.
- Added ped & object spawner.
- Moved vehicle related things to a separate submenu.


lonelybud:
does the call mechanic feature reset cooldown as well?

ShinyWasabi:
> does the call mechanic feature reset cooldown as well?

It should now.

PR #332, opened 2025-05-22
feat(vehicle): Fix All Personal Vehicles

ShinyWasabi:


src/types/stats/CStatsMpCharacterMappingData.cpp
```diff
@@ -0,0 +1,23 @@
+#include "CStatsMpCharacterMappingData.hpp"
+
+int CStatsMpCharacterMappingData::GetStatCount(Type type)
```
maybegreat48, line 3: Are we using this for anything right now or is it for later?

ShinyWasabi: I will use it to iterate through stats when I add an award editor.

PR #340, opened 2025-05-25
feat(Teleport): add Teleport to Taskpoint

Cleanwn:


xesdoog:
Objectives have several blip IDs, colors, states, etc.. `ID 1` is a basic circle blip that's used for a lot of things, including players (just a circle blip with a heading indicator). This means that this code will end up teleporting you to [the nearest "circle" blip](https://github.com/YimMenu/YimMenuV2/pull/340/files#diff-6163caee0bb8a5f32b77ea9ab83c844f51ebb6c83a45cbb409bae00b97cc0e9eR92), which can be another player or something else that's not an objective.

There's another [Objective TP PR](https://github.com/YimMenu/YimMenuV2/pull/263) and although it iterates through multiple blip IDs instead of just `1`, it is still problematic. This was an issue in legacy YimMenu as well and it was fixed by gir489returns. Sicne their account was terminated, I was unable to find the PR in the archived YimMenu repository but this is everything related to blips and objective teleport from legacy *(links point to @Mr-X-GTA 's repos so if this is not allowed, please lmk)*:

- [Classes](https://github.com/Mr-X-GTA/GTAV-Classes-1/tree/master/ui)
- [blip_util](https://github.com/Mr-X-GTA/YimMenu/blob/master/src/util/blip.cpp#L51)

PR #341, opened 2025-05-27
ci: add Clang build

eXhumer:
* add Clang builds alongside MSVC
* include all build types (Release, Debug, RelWithDebInfo & MinSizeRel)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,12 +1,11 @@
-# TODO: use Clang instead of MSVC
 name: CI
 
 on: [ push, pull_request ]
 
 jobs:
-  ci:
+  msvc-ci:
     runs-on: [ self-hosted, Windows ]
-    name: CI
+    name: CI (MSVC)
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


maybegreat48:
This will cause the menu to build 8 times whenever a commit is pushed. Is that really necessary? I would just keep a clang build and an MSVC build

GCC on Linux is useful too, if you get it to work

eXhumer:
I can remove any build type you feel is un-necessary. The nightly builds are RelWithDebInfo and the CI builds are currently Release.

tupoy-ya:
> GCC on Linux is useful too, if you get it to work

Pretty sure you can't call virtual functions with it, so it's not very useful.

maybegreat48:
> Pretty sure you can't call virtual functions with it, so it's not very useful.

Interesting. I always thought (early versions of YimMenuV2) could compile on Linux and work fine

> I can remove any build type you feel is un-necessary. The nightly builds are RelWithDebInfo and the CI builds are currently Release.

I would use `RelWithDebInfo` for both build types. The CI build should include MSVC, Clang, and (maybe) GCC

.github/workflows/ci.yml
```diff
@@ -19,10 +18,10 @@ jobs:
           arch: amd64
 
       - name: Generate CMake project
-        run: cmake -D CMAKE_BUILD_TYPE=Release -S. -Bbuild -G Ninja
+        run: cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DOPTIMIZE=YES -S . -B build -G Ninja
```
tupoy-ya: What does `-DOPTIMIZE=YES` do?
It's not used anywhere and cmake doesn't do anything with it internally.

eXhumer: Nightly builds use it. I'm just trying to mimic the nightly build flags here. I'll remove it if it does nothing.

eXhumer: For reference

https://github.com/YimMenu/YimMenuV2/blob/15a74630f95a7ecc9abf05d7b3ea5d5cb4eb0b09/.github/workflows/nightly.yml#L53

tupoy-ya: It was added in YimV1 here: https://github.com/YimMenu/YimMenu/pull/2818
and used here:
https://github.com/YimMenu/YimMenu/blob/24ae18cad2a3e2d8a68fc01e58dd3bbf982c48a8/CMakeLists.txt#L79-L88

YimMenuV2 does not use this. It was probably copied into nightly builds from YimMenuV1.

eXhumer: removed with ebfd0d9

PR #343, opened 2025-05-27
ci: replace MSVC with Clang for builds

eXhumer:


eXhumer:
Is there any specific reason why self-hosting is being used instead of public runners?

Yimura:
We ran out of memory on YimMenu `v1`, as a solution we swapped to using a self-hosted Windows Github Runner to handle builds under the YimMenu organisation.

I manage the self-hosted instance and was not aware of a `clang` build variant suddenly being used.
The reason for failing builds is that each and every tool needs to be manually setup on a self-hosted runner as opposed to the Github provided Windows runners.

As for the previous PR and this PR, _why_ are we making use of clang next to msvc to compile V2?
The _why_ reasoning was missing from the original PR and it feels a bit pointless wasting additional resources on this.,

eXhumer:
For the reason of adding Clang builds, it was due to the following comment I noticed on CI workflow file.

https://github.com/YimMenu/YimMenuV2/blob/8c74b2b801a0f887ed546ea671515e0fa1f2a327/.github/workflows/ci.yml#L1

And, as for the reason of running both MSVC and Clang jobs, I feel it's best to provide both for a few CI releases until it is certain there is no issues with Clang builds before removing the MSVC builds.

maybegreat48:
>  The why reasoning was missing from the original PR and it feels a bit pointless wasting additional resources on this.,

GTA uses Clang to build its binaries now. I haven't discovered any ABI breaks when using MSVC yet, but compiling the menu with Clang should still provide a stabler experience

Again, if you don't want to, I can disable the Clang (or MSVC) build

Yimura:
In that case it's fine by me, I was wondering why exactly.

maybegreat48:
@eXhumer can we just remove the MSVC build? And swap the nightly builds to use Clang?

eXhumer:
Don't merge yet, let me verify all actions work correctly.

eXhumer:
@Yimura Can you install LLVM on the self-hosted machine and add it to PATH.

The action I found to install LLVM require 7z to be added to PATH, and trying to install 7z on Windows require manual confirmation on machine as the installation binary is not signed by the 7z developer.

This way I can avoid having to install / set it up manually.

Or, I have an alternative solution of just using the public runner instead, where all the necessary binaries already provided by default. This shouldn't cause problems with building as I don't believe YimMenuV2 have memory issues, atleast when I tested on my fork last time.

Yimura:
I will install it at some point, just haven't gotten around to doing it.

Yimura:
I've installed LLVM, added the `bin/` directory to `PATH` and rebooted the machine to be sure the runner has the new `PATH` variable.

eXhumer:
@Yimura @maybegreat48 Ready to merge.

CI Build Run: https://github.com/eXhumer/YimMenuV2/actions/runs/15367754438
Nightly Build Run: https://github.com/eXhumer/YimMenuV2/actions/runs/15367866334

eXhumer:
Ignore the abysmally long build times, it was due to me having to build it on a Macbook using Parallels.

Yimura:
Ninja already exists on the machine, there's no need to download it.

Yimura:
@maybegreat48 I leave the rest to you, unsure what else is required before this can be merged (testing if it'll actually run for example).

eXhumer:
For what it is worth, I have verified and tested the files to be working without issue, but I assume you would want to do independent testing to verify this yourself.

.github/workflows/ci.yml
```diff
@@ -3,9 +3,9 @@ name: CI
 on: [ push, pull_request ]
 
 jobs:
-  msvc-ci:
+  ci:
     runs-on: [ self-hosted, Windows ]
-    name: CI (MSVC)
+    name: CI
     steps:
       - uses: actions/checkout@v3
```
CanerKaraca23, line 10:
```suggestion
      - uses: actions/checkout@v4
```

eXhumer: Is there a reason for this? v3 has not been deprecated as far as I am aware.

eXhumer:
@maybegreat48 any change needed or encountered any issue? hoping to see it merged otherwise.

.github/workflows/ci.yml
CanerKaraca23: > Is there a reason for this? v3 has not been deprecated as far as I am aware.

It's deprecated, v4 is latest.

eXhumer: Show me the deprecation notice

CanerKaraca23: > Show me the deprecation notice

v3 uses node 16, it's deprecated on GitHub, so... Why not using latest?

PR #346, opened 2025-05-27
fix: aimbot not switching to new targets

felixwehrfeld:
Already tested, working fine.


src/game/features/self/Aimbot.cpp
```diff
@@ -1,18 +1,42 @@
...
 namespace YimMenu::Features
 {
-	class Aimbot : public BoolCommand
+	class Aimbot : public LoopedCommand
 	{
-		using BoolCommand::BoolCommand;
+		bool running = false;
+		using LoopedCommand::LoopedCommand;
+
+		std::set<int> killed_peds = {  };
```
maybegreat48, line 17: This'll eventually fill up since it's never cleared

src/game/features/self/Aimbot.cpp
```diff
@@ -1,18 +1,42 @@
...
 
 		virtual void OnEnable() override
 		{
 			Pointers.ShouldNotTargetEntityPatch->Apply();
 			Pointers.GetAssistedAimTypePatch->Apply();
 		}
 
+		virtual void OnTick() override
+		{
+			if (running) return;
```
maybegreat48, line 27: You don't need the running check since `OnTick` isn't called from multiple threads

src/game/features/self/Aimbot.cpp
felixwehrfeld: fixed at https://github.com/YimMenu/YimMenuV2/pull/346/commits/e84a26765d9cbf8004ed09c005c1406d71501ff2

src/game/features/self/Aimbot.cpp
felixwehrfeld: @maybegreat48 this is necessary to prevent the game from freezing, I don't know why, but it seems like the game likely freezes if the loop stacks because it is maybe slower than the OnTick(). Any suggestions?

maybegreat48: That's... very unusual. I'll look into it, thanks for letting me know

felixwehrfeld: @maybegreat48 any news so we can merge soon?

felixwehrfeld:
@maybegreat48 any news?

maybegreat48:
I'm trying to investigate better ways to do this

felixwehrfeld:
@maybegreat48 I also searched for a better way, but I found nothing. This is also the kind of way the common MP-Client-Events work.
Can u merge it either? If there is a better way, we can improve it but it would be nice for the people if they can use the aimbot this way till we found a better solution.

maybegreat48:
This has been fixed through a hook

eXhumer:
@maybegreat48 has it? I tested it today, while it no longer stays locked to a killed ped, it still doesn't switch to new target on area with clear line of sight.

maybegreat48:
Should be fixed now

PR #353, opened 2025-05-31
Fix: Aimbot Simplified logic

XxRagulxX:
- Removed unnecessary running flag since OnTick() is called sequentially by the command system.
- Added **killed_peds.clear()** in both OnEnable() and OnDisable() to properly reset state when toggling the feature.
- Improved code clarity by consistently storing and using handle instead of multiple calls to **ped.GetHandle()**.


maybegreat48:
This is just a low quality AI edit of #346

PR #374, opened 2025-06-06
add Vehicle Editor & persist car feature

lonelybud:
add Vehicle Editor
- add VehicleEditorCategory in Vehicle Submenus
- add GetOwnedMods, GetVehicleFullname in Vehicle class
- add VehicleColorsChameleon in VehicleValues
- add VehicleHelper core/util
- add lscTypes in gta/data

add persist car
- add GetFile function in Folder class
- add core/util/strings.hpp
- add PersistCarService in features
- add PersistCarCategory in Vehicle Submenus

Features are copied from YimMenu Legacy.

PR #376, opened 2025-06-06
added some LSC Features

CSY0N:
- Updated main.cpp
- Updated Vehicle.hpp
- Added VehicleLSCCategory.hpp
- Added VehiclePersistCarCategory.hpp


tupoy-ya:
The code in this and in #374 is pretty similar, they even have the same comments.
Except this is lower quality and tries to break to the menu by removing random things from main.cpp (why?)

Does this even compile?

lonelybud:
lol dont copy the code from my fork XD. wat are you doing mate

> Does this even compile?

it will not. Its straight away copied without even thinking.

abuazizv:
This is just AI garbage.

PR #388, opened 2025-06-07
feat: add custom weapons

kurumimeow:
Added custom weapons (port from YimMenu):

- Cage Gun
- Delete Gun
- Gravity Gun
- Steal Vehicle Gun
- Repair Gun
- Vehicle Gun
- Teleport Gun
- Paint Gun

Added more UI Items:

- TabBarItem


CSY0N:
your Code make conflicts with the base fix it

kurumimeow:
> your Code make conflicts with the base fix it

It's resolved.

Deadlineem:
> > your Code make conflicts with the base fix it
> 
> It's resolved.

Merge conflicts again with latest.  @maybegreat48 can we add this?  I was going to do this myself but seeing a PR for it, no need.

@kurumimeow as for the SameLineItem, i think this might cause issues since each "section" in the menu has a limited height factor, when that is reached, the items dynamically move to the right side of the existing options.  You can see an example of this in the current build by going to Self > Weapons and toggling all 3 of these on

Explosive Ammo
Weapon Damage
Melee Damage

The float inputs and options move themselves.  Double check to ensure this doesn't cause issues with how it works currently after your merge

maybegreat48:
Yeah, I don't recommend creating something like SameLineItem

kurumimeow:
The SameLineItem will not cause any issue, it actually removes the junkies from the code. It is really useful if you want some items in different line but some items in same line in a same group.
`loots->AddItem(std::make_shared<ImGuiItem>([] {
	ImGui::SameLine();
}));`
`loots->AddItem(std::make_shared<CommandItem>("cayopericoheistsetprimarytargetvalue"_J, "Set##primarytargetvalue"));`
As an example for this one, it will count as two items, but if you use SameLineItem it counts as one item.

And the group takes 7 items per column by default (the items_per_row variable is actually items_per_column here). If you want Explosive Ammo, Weapon Damage, Melee Damage in different line you can remove the second parameter or replace with 3. https://github.com/YimMenu/YimMenuV2/blob/51f5d859eb3dc4e53a7f52e26e6b6efe41084c49/src/game/frontend/items/Items.hpp#L141

maybegreat48:
You can just keep the items you want on the same line with an empty group by passing an empty string as the name parameter

kurumimeow:
Yeah, that can be done too, I'll remove the SameLineItem then.

CSY0N:
you should fix ur code not works correctly.

kurumimeow:
@maybegreat48 i have removed the SameLineItem

CSY0N:
> @maybegreat48 i have removed the SameLineItem

**### ERROR** has been find in **_CostumWeapon.cpp_**

this line
- `auto veh = Vehicle::Create(modelHash, spawnLoc, Self::GetPed().GetHeading(), false);`

change it to this: `auto veh = Vehicle::Create(modelHash, spawnLoc, Self::GetPed().GetHeading());`

CSY0N:
will this be added or should i fix it? that's get working without problem?

src/core/util/RainbowColor.hpp
```diff
@@ -0,0 +1,63 @@
+#pragma once
+#include "types/script/Timer.hpp"
```
maybegreat48: You shouldn't be using the script timer for this stuff, nor a script vector. `std::chrono::system_clock` is a good alternative, and create a color struct instead of using a vector

src/game/features/self/CustomWeapon.cpp
```diff
@@ -0,0 +1,438 @@
...
+							});
+						}
+
+						vehicleGunTimer.Reset(true);
+					}
+				}
+			}
+			break;
+		}
+		case CustomWeapons::TELEPORT_GUN:
```
maybegreat48, line 304: This should also set you in any vehicle you shoot at

src/game/features/self/CustomWeapon.cpp
```diff
@@ -0,0 +1,438 @@
...
+					velocity.y *= (float)multiplier;
+					velocity.z *= (float)multiplier;
+
+					entity.SetVelocity(velocity);
+					entity.SetAlpha(105);
+				}
+			};
+
+			FiberPool::Push([=]() mutable {
+				if (isCustomWeaponActionAllowed && PAD::IS_DISABLED_CONTROL_PRESSED(0, (int)ControllerInputs::INPUT_AIM))
```
maybegreat48: There's a lot of code duplication with the customWeaponAction code that must be refactored into something else if possible

src/game/features/self/CustomWeapon.cpp
```diff
@@ -0,0 +1,438 @@
...
+#include "types/script/Timer.hpp"
+
+namespace YimMenu::Features
+{
+	void CustomWeapon::OnTick()
+	{
+		CustomWeapons selectedWeaponType = static_cast<CustomWeapons>(_CustomWeaponType.GetState());
+
+		if (bool isGravityGunSelected = (selectedWeaponType == CustomWeapons::GRAVITY_GUN); _IsGravityGunEnabled.GetState() != isGravityGunSelected)
+			_IsGravityGunEnabled.SetState(isGravityGunSelected);
```
maybegreat48: I see why you're doing this, but you might want to overload/duplicate ConditionalItem with something that'll accept arbitrary callbacks so you can check whether the list item state matches `GRAVITY_GUN` etc. This will also be useful for future features that require a more complicated availability check than just a BoolCommand being enabled

src/game/features/self/CustomWeapon.cpp
```diff
@@ -0,0 +1,438 @@
...
+									auto location = m_Entity.GetPosition();
+									auto distance = Self::GetPed().GetPosition().GetDistance(location);
+
+									if (distance > 500)
+									{
+										Notifications::Show("Custom Weapon", "Entity is too far.", NotificationType::Warning);
+									}
+									else
+									{
+										if (m_Entity.RequestControl(300))
```
maybegreat48: You don't need control to delete objects, including cloned ones

src/game/features/self/CustomWeapon.cpp
```diff
@@ -0,0 +1,438 @@
...
+		}
+		case CustomWeapons::DELETE_GUN:
+		{
+			if (isCustomWeaponActionAllowed)
+			{
+				if (PAD::IS_DISABLED_CONTROL_PRESSED(0, (int)ControllerInputs::INPUT_AIM))
+				{
+					if (PAD::IS_DISABLED_CONTROL_JUST_RELEASED(0, (int)ControllerInputs::INPUT_ATTACK))
+					{
+						FiberPool::Push([=]() mutable {
```
maybegreat48: Nor do you need a fiber pool

src/game/features/self/CustomWeapon.cpp
```diff
@@ -0,0 +1,438 @@
...
+										else
+										{
+											Notifications::Show("Custom Weapon", "Failed to take control of entity.", NotificationType::Error);
+										}
+									}
+								}
+							}
+							else
+							{
+								Notifications::Show("Custom Weapon", "No entity found.", NotificationType::Warning);
```
maybegreat48: I would remove this notification since it'll get pretty annoying

src/game/features/self/CustomWeapon.cpp
```diff
@@ -0,0 +1,438 @@
...
+							if (RayCast(m_Entity))
+							{
+								auto entityHandle = m_Entity.GetHandle();
+
+								if (m_Entity.IsVehicle())
+								{
+									for (size_t i = 0; i < 8 && !m_Entity.As<Vehicle>().IsSeatFree(-1); i++)
+									{
+										const auto ped = VEHICLE::GET_PED_IN_VEHICLE_SEAT(entityHandle, -1, 0);
+										TASK::CLEAR_PED_TASKS_IMMEDIATELY(ped);
```
maybegreat48: This doesn't work with player peds anymore, so you may want to add a check for that and abort. You also don't need to wait after kicking the ped out, just immediately set yourself in the vehicle and it'll work

src/game/features/self/CustomWeapon.cpp
```diff
@@ -0,0 +1,438 @@
...
+			if (isCustomWeaponActionAllowed)
+			{
+				if (PAD::IS_DISABLED_CONTROL_PRESSED(0, (int)ControllerInputs::INPUT_AIM))
+				{
+					if (PAD::IS_DISABLED_CONTROL_PRESSED(0, (int)ControllerInputs::INPUT_ATTACK))
+					{
+						Vector3 coords;
+
+						FiberPool::Push([=]() mutable {
+							if (RayCast(coords))
```
maybegreat48: You don't need to call `ADD_DECAL` in a fiber pool

src/game/features/self/CustomWeapon.cpp
```diff
@@ -0,0 +1,438 @@
...
+					if (PAD::IS_DISABLED_CONTROL_JUST_RELEASED(0, (int)ControllerInputs::INPUT_ATTACK))
+					{
+						if (RayCast(m_Entity))
+						{
+							if (m_Entity.IsPed())
+							{
+								auto location = m_Entity.GetPosition();
+								location.z -= 1.0f;
+
+								Object::Create("prop_gold_cont_01"_J, location);
```
maybegreat48, line 43: This one needs to be in a fiber pool since it may yield

src/game/features/self/CustomWeapon.hpp
```diff
@@ -0,0 +1,139 @@
...
+		ControllerInputs::INPUT_ATTACK,
+		ControllerInputs::INPUT_ATTACK2,
+		ControllerInputs::INPUT_VEH_ATTACK,
+		ControllerInputs::INPUT_VEH_ATTACK2,
+		ControllerInputs::INPUT_VEH_PASSENGER_ATTACK,
+		ControllerInputs::INPUT_VEH_FLY_ATTACK,
+		ControllerInputs::INPUT_VEH_FLY_ATTACK2,
+	};
+
+	static bool RayCast(Entity& entity)
```
maybegreat48: The raycast functions should be moved somewhere else (`game/gta/Raycast.hpp`?) You may also wish to combine both functions to get the hit entity and the position at the same time

src/game/gta/data/DecalTypes.hpp
```diff
@@ -0,0 +1,57 @@
+#pragma once
+
+enum class DecalTypes
```
maybegreat48: This should be in `types/entity/DecalTypes.hpp`

kurumimeow:
Done

PR #400, opened 2025-06-09
fix(superrun): use native methods to modify speed

eXhumer:
* Modify speed using the [`SET_PED_MOVE_RATE_OVERRIDE`](https://docs.fivem.net/natives/?_0x085BF80FA50A39D1) native method instead of calculating speed manually and setting it. This essentially is a multiplier to the base speed. This fixes backwards movement in first-person mode and while using stairs compared to previous.
* Modify run / sprint & swim speed multiplier using [`SET_SWIM_MULTIPLIER_FOR_PLAYER`](https://docs.fivem.net/natives/?_0xA91C6F0FF7D16A13) & [`SET_RUN_SPRINT_MULTIPLIER_FOR_PLAYER`](https://docs.fivem.net/natives/?_0x6DB47AA77FD94E09) native methods respectively.

### DIFF
--- a/src/game/frontend/submenus/Vehicle/SavedVehicles.cpp
+++ b/src/game/frontend/submenus/Vehicle/SavedVehicles.cpp
@@ -2,7 +2,7 @@
 
 #include "core/backend/FiberPool.hpp"
 #include "core/frontend/Notifications.hpp"
-#include "core/util/Strings.hpp"
+#include "core/util/strings.hpp"
 #include "game/backend/Self.hpp"
 #include "game/backend/SavedVehicles.hpp"
 #include "game/gta/Vehicle.hpp"
--- a/src/game/features/self/SuperRun.cpp
+++ b/src/game/features/self/SuperRun.cpp
@@ -1,57 +1,43 @@
 #include "core/commands/LoopedCommand.hpp"
 #include "core/commands/FloatCommand.hpp"
-#include "core/util/Math.hpp"
 #include "game/backend/Self.hpp"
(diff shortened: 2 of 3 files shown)


eXhumer:
The reason why I didn't really set the correct upper limit for the options is because if I do that, the GUI replace it with a slider instead. Do you think an option to override the slider option is worth adding?

src/game/frontend/submenus/Vehicle/SavedVehicles.cpp
```diff
@@ -2,7 +2,7 @@
 
 #include "core/backend/FiberPool.hpp"
 #include "core/frontend/Notifications.hpp"
-#include "core/util/Strings.hpp"
+#include "core/util/strings.hpp"
```
maybegreat48, line 5: Ah interesting. It seems renaming the file locally (Strings.hpp -> strings.hpp) doesn't affect Git

eXhumer: You can use `git mv` if you want to rename it.

maybegreat48:
> The reason why I didn't really set the correct upper limit for the options is because if I do that, the GUI replace it with a slider instead. Do you think an option to override the slider option is worth adding?

Is there a reason we don't want the slider?

eXhumer:
> Is there a reason we don't want the slider?

For floats, precise value control. For the run/sprint & swim multiplier, the range of value is between 1.0 and 1.49. It would be better to have precise control over the exact value used rather than whatever step ImGui thinks is appropriate as a step.

PR #403, opened 2025-06-10
feat(ui): allow configuration of slider for int/float command items

eXhumer:
See https://github.com/YimMenu/YimMenuV2/pull/400#issuecomment-2957788866 for more information.

This will keep all existing slider as is, only will explicitly disable it when `use_slider` argument to `false`.

### DIFF
--- a/src/game/frontend/items/FloatCommandItem.cpp
+++ b/src/game/frontend/items/FloatCommandItem.cpp
@@ -6,7 +6,8 @@
 
 namespace YimMenu
 {
-	FloatCommandItem::FloatCommandItem(joaat_t id, std::optional<std::string> label_override) :
+	FloatCommandItem::FloatCommandItem(joaat_t id, std::optional<std::string> label_override, bool use_slider) :
+	    m_useSlider(use_slider),
 	    m_Command(Commands::GetCommand<FloatCommand>(id)),
 	    m_LabelOverride(label_override)
 	{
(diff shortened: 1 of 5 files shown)


src/game/frontend/items/FloatCommandItem.cpp
```diff
@@ -6,7 +6,8 @@
 
 namespace YimMenu
 {
-	FloatCommandItem::FloatCommandItem(joaat_t id, std::optional<std::string> label_override) :
+	FloatCommandItem::FloatCommandItem(joaat_t id, std::optional<std::string> label_override, bool use_slider) :
+		m_useSlider(use_slider),
```
eXhumer: The reason this looks messed up is because I am using tabs as spacing exclusively, while the lines below seem to have a combination of tabs and spaces.

src/game/frontend/items/IntCommandItem.cpp
```diff
@@ -6,7 +6,8 @@
 
 namespace YimMenu
 {
-	IntCommandItem::IntCommandItem(joaat_t id, std::optional<std::string> label_override) :
+	IntCommandItem::IntCommandItem(joaat_t id, std::optional<std::string> label_override, bool use_slider) :
+		m_useSlider(use_slider),
```
eXhumer: Same as above.

eXhumer:
The project should have a stricter `.clang-format` and should reject PRs if styling check is failed. Maybe I'll try to see if it is feasible to add clang-format step in CI to automate this...

src/game/frontend/items/FloatCommandItem.cpp
```diff
@@ -6,7 +6,8 @@
 
 namespace YimMenu
 {
-	FloatCommandItem::FloatCommandItem(joaat_t id, std::optional<std::string> label_override) :
```
maybegreat48, line 9: Just add a step value and make it not use the slider if it's 0/0,0f

eXhumer: ImGui doesn't support custom steps natively.

I'll look at ImGui to see what the best way to implement it would be.

PR #405, opened 2025-06-10
feat(recovery): free change appearance & allow gender change

djandre00:


src/game/features/recovery/AllowGenderChange.cpp
```diff
@@ -0,0 +1,22 @@
...
+		using BoolCommand::BoolCommand;
+
+		virtual void OnEnable() override
+		{
+			Stats::SetInt("MPx_ALLOW_GENDER_CHANGE", 52);
+		}
+
+		virtual void OnDisable() override
+		{
+			Stats::SetInt("MPX_HEIST_PLANNING_STAGE", 0);
```
eXhumer: Is the correct? Shouldn't you be changing `MPx_ALLOW_GENDER_CHANGE` instead.

djandre00: yes my bad, its fixed now

PR #407, opened 2025-06-10
lint: lint source & add ci to lint check before building

eXhumer:


eXhumer:
The lint job will be running using GitHub hosted runner as this doesn't need to be run on the self-hosted.

maybegreat48:
So does this mean any lint failures will halt the CI build? The clang-format file is basically a guideline at this point. not a rule that all files have to follow. I sometimes intentionally stray from clang's formatting to improve readability, and some formatting options (such as reordering header files) might even affect compilation output

Clang-tidy is more useful, but I never managed to set that up correctly

maybegreat48:
I've made the clang-format file a little more bearable. Can you reformat the repository so I can see if this'll work?

eXhumer:
If you prefer not to block CI even if lint fails, I can set that up as well.

eXhumer:
https://github.com/eXhumer/YimMenuV2/actions/runs/15585553513

Here is an example run where lint fails, but CI runs as usual.

eXhumer:
@maybegreat48 just be aware the current implementation will block CI job if lint fails. if you don't want CI to depend on lint, simply remove line `needs: lint` from the ci job.

PR #450, opened 2025-06-17
updated patterns & Globals.cpp

SourceModzZ:


Deadlineem:
@SourceModzZ 

constexpr auto handleJoinRequestIgnorePoolPatchPtrn = Pattern<"83 FD 05 ? ? ? 00 00 00 48 8B">("HandleJoinRequestIgnorePoolPatch");
scanner.Add(handleJoinRequestIgnorePoolPatchPtrn, [this](PointerCalculator ptr) {
	HandleJoinRequestIgnorePoolPatch = BytePatches::Add(ptr.Add(3).As<std::uint8_t*>(), std::to_array<std::uint8_t>({0x48, 0xE9}));
});

This was commented on my issue topic by @tupoy-ya and since there is a PR for part of the problem already, may as well add it.

SourceModzZ:
> @SourceModzZ
> 
> constexpr auto handleJoinRequestIgnorePoolPatchPtrn = Pattern<"83 FD 05 ? ? ? 00 00 00 48 8B">("HandleJoinRequestIgnorePoolPatch"); scanner.Add(handleJoinRequestIgnorePoolPatchPtrn, [this](PointerCalculator ptr) { HandleJoinRequestIgnorePoolPatch = BytePatches::Add(ptr.Add(3).Asstd::uint8_t*(), std::to_arraystd::uint8_t({0x48, 0xE9})); });
> 
> This was commented on my issue topic by @tupoy-ya and since there is a PR for part of the problem already, may as well add it.

Done

src/game/features/self/DisablePhone.cpp
```diff
@@ -0,0 +1,22 @@
+#include "core/commands/LoopedCommand.hpp"
+#include "game/gta/ScriptGlobal.hpp"
+
+namespace YimMenu::Features
```
maybegreat48: I think you forgot to remove this

SourceModzZ: Done 👍

Deadlineem: > I think you forgot to remove this

Out of curiosity, don't they both do different things?

One silences the sounds but allows the phone to pop up still, the other i believe disables the phone completely, right?  So in all actuality, he could probably implement this in the same file as silent phone calls and make it conditional so if silent phone calls is active, the option will show to disable the phone itself?

maybegreat48: People generally disable the phone to avoid the annoying phone calls, not because they don't want the phone

PR #452, opened 2025-06-17
Added disablephone to silencephonecalls & New DLC Vehicles

SourceModzZ:


src/game/features/network/SilencePhonecalls.cpp
```diff
@@ -27,5 +27,30 @@ namespace YimMenu::Features
...
+			if (*Pointers.IsSessionStarted)
+			{
+				constexpr auto prevent_phone = ScriptGlobal(21222);
+				*prevent_phone.As<int*>() = 0;
+			}
+		}
+	};
+
 	static SilencePhonecalls _SilencePhonecalls{"nocalls", "Silence Phone Calls", "Automatically silences all incoming calls"};
+	static PreventPhonePopup _PreventPhonePopup{"disablephone", "Prevent Phone Popups", "Prevents the phone from showing up on incoming calls"};
```
maybegreat48: What does this do? Does this block the phone completely? The phone already doesn't show up when the above option is enabled

Deadlineem: > What does this do? Does this block the phone completely? The phone already doesn't show up when the above option is enabled

Game still crashes for me after loading into story mode, applied all your patches and cout.log is clear from errors.

Are we 100% sure the current Silence calls option disables the phone from popping up, too?  I noticed the global for his disabler is not one used in silencecalls

SourceModzZ: I give @Deadlineem Right

Deadlineem: > What does this do? Does this block the phone completely? The phone already doesn't show up when the above option is enabled

Update on this, idk if its bc of the update or bc im in story mode but the silence calls options is not silencing calls nor preventing the phone from popping up.
Screenshot_126

SourceModzZ: one sec. Let me fix it

maybegreat48: Yeah, I made it MP only because it'll probably break some missions in SP

SourceModzZ: > Yeah, I made it MP only because it'll probably break some missions in SP



> Yeah, I made it MP only because it'll probably break some missions in SP

I have updated the globals

src/game/features/network/SilencePhonecalls.cpp
```diff
@@ -27,5 +27,30 @@ namespace YimMenu::Features
 		}
 	};
 
+	class PreventPhonePopup : public LoopedCommand
```
maybegreat48: Can you remove this feature?

SourceModzZ: Done

PR #480, opened 2025-06-18
Add duplicate checkbox to issue templates

Yimura:
Add a checkbox to the issue template in the hopes that users spend at least a few more seconds verifying if their issue is a duplicate or not.


maybegreat48:
People will just check the box and move on, but hey, worth a shot

PR #485, opened 2025-06-18
(Improvement) Melee, Explosive, Weapon damage

Deadlineem:


Deadlineem:
NOTE: The melee damage also works with ALL melee based weapons including bare fists.  Melee damage DOES NOT apply when holding a gun and butting the target with the gun, if you want that damage effect when gun butting enemies, use Weapon damage :)

PR #492, opened 2025-06-18
Fixed CayoPericoHeist & DoomsdayHeist

SourceModzZ:
I had forgotten to update a global correctly now everything is correct


AnikSingha12:
um the casino heist wasnt working for me , i setup from mod menu and then when going to finale game crashes, i tried doing the setup myself and then going to cayo but still crashes, then i use battleeye to go to public lobby and then i found that the weapon mission i did normally isnt registered, so any fix?

PR #503, opened 2025-06-19
(Improvement) Custom UI Styles

Deadlineem:
Easily add new UI styles by creating a folder in the following location and adding the Custom style .cpp inside of that folder:
`core/frontend/manager/styles`

Example:
<folder>MyCustomStyle/<file>MyCustomStyle.cpp

- Automatic listings of new compiled styles show your custom style in the listbox without needing to add menu options to Settings.cpp

https://github.com/user-attachments/assets/49c5985c-a0d0-4959-b8f4-f51d203a9adf

**Current working styles are:**
Original (Classic) style 
Modern (New UI design I created)


Deadlineem:
https://github.com/user-attachments/assets/318480bf-cd2a-40bb-bc14-478d1c6c442a

L7NEG:
Big W fr

src/core/frontend/manager/UIManager.hpp
```diff
@@ -42,7 +42,7 @@ namespace YimMenu
 			return GetInstance().m_ShowContentWindow;
 		}
 
-	private:
+	public:
```
maybegreat48: This should not be public

src/game/frontend/submenus/Settings/StyleSelector.cpp
```diff
@@ -0,0 +1,49 @@
+﻿#include "StyleSelector.hpp"
+#include "core/commands/ListCommand.hpp"
+#include "core/frontend/manager/styles/StyleManager.hpp"
+
+namespace YimMenu
+{
+	// Build style options statically
+	static const std::vector<std::pair<int, const char*>> g_StyleOptions = [] {
```
maybegreat48: You're relying on this running after the registrar functions, which is [undefined behavior](https://en.cppreference.com/w/cpp/language/siof.html). Just get rid of the registrar and manually assign each style with an index in UIManager.hpp and use a switch statement to call into the different themes. You can also get rid of StyleManager after this

src/core/frontend/manager/styles/Classic/Classic.cpp
```diff
@@ -0,0 +1,81 @@
+#include "core/frontend/manager/styles/StyleManager.hpp"
```
maybegreat48: I don't think we need a directory for each style

src/core/frontend/manager/UIManager.hpp
```diff
@@ -54,10 +54,27 @@ namespace YimMenu
 		void DrawImpl();
 		std::shared_ptr<Submenu> GetActiveSubmenuImpl();
 		std::shared_ptr<Category> GetActiveCategoryImpl();
-
 		std::shared_ptr<Submenu> m_ActiveSubmenu;
 		std::vector<std::shared_ptr<Submenu>> m_Submenus;
+
 		ImFont* m_OptionsFont = nullptr;
 		bool m_ShowContentWindow = true;
+		bool g_StyleWasJustRestored = false;
```
maybegreat48: What does this do?

src/core/frontend/manager/UIManager.hpp
```diff
@@ -54,10 +54,27 @@ namespace YimMenu
...
+		{
+			GetInstance().m_ShowContentWindow = show;
+		}
+
+		static const std::vector<std::shared_ptr<Submenu>>& GetSubmenus()
+		{
+			return GetInstance().m_Submenus;
+		}
+
+		ImFont* GetOptionsFont()
```
maybegreat48: Should be static and should `return GetInstance().m_OptionsFont;`

src/core/frontend/manager/UIManager.hpp
```diff
@@ -54,10 +54,27 @@ namespace YimMenu
...
 		std::shared_ptr<Category> GetActiveCategoryImpl();
-
 		std::shared_ptr<Submenu> m_ActiveSubmenu;
 		std::vector<std::shared_ptr<Submenu>> m_Submenus;
+
 		ImFont* m_OptionsFont = nullptr;
 		bool m_ShowContentWindow = true;
+		bool g_StyleWasJustRestored = false;
+
+		static void SetShowContentWindow(bool show)
```
maybegreat48: If you want these functions to be accessible to things outside this class place them right after `ShowingContentWindow`

src/core/frontend/manager/styles/Classic/Classic.cpp
Deadlineem: > I don't think we need a directory for each style

The directory is more for future plans, like maybe the style could have different color presets.  For example, Classic folder could have Classic.cpp and Classic_Monotone.cpp and when classic is selected it will apply and then on the right of the style listbox, or below it, will show the preset color variants.  

This is more or less worked around the Modern theme or any theme which has its own color options instead of using the SetupStyle in Menu cpp

This could also be more useful to keep folders if you wanted to have multiple files for one theme (Main.cpp, buttons.cpp, colors cpp, etc.) however could probably have the theme folders inside of a themes directory so its not bulking up the styles folder -see styleselector comments-

src/core/frontend/manager/UIManager.hpp
Deadlineem: It was something to attempt fixing the Classic UI i forgot to take out when i got it working, thanks for catching that

src/game/frontend/submenus/Settings/StyleSelector.cpp
Deadlineem: Thank you for this, i had a hell of a time last night, anywho i would be happy to get rid of the need for style manager because that will leave the styles folder for storing all the custom style folders -see comment about style folders-

src/core/frontend/manager/UIManager.hpp
Deadlineem: For some reason it was not allowing me to utilize the variables when they were private, i will try again and test, maybe after all these changes it will resolve my problems, most of this was assistance based i struggled a bit with this PR lmfao

src/core/frontend/manager/UIManager.hpp
Deadlineem: This was also another variable i wasnt able to utilize and it drove me nuts 😂 anything listed public in the hpp now was causing me hella trouble.

src/core/frontend/manager/UIManager.hpp
Deadlineem: Thanks, will make all the changes, ill dm if i run into trouble

src/core/frontend/manager/styles/ModernVertical/ModernVertical.cpp
```diff
@@ -0,0 +1,159 @@
+#include "game/pointers/Pointers.hpp"
+#include "game/frontend/Menu.hpp"
+#include "core/frontend/manager/UIManager.hpp"
+
+namespace YimMenu
+{
+	void ApplyModernVTheme()
```
maybegreat48: No I mean something like creating a header file for each theme and exposing a `RenderModernVTheme` etc then importing all these headers at UIManager.cpp and then calling the funcs

Deadlineem: Oh man, I was going to make a Themes.hpp to store them all, should I just do that instead of creating a header file for each individual theme?

src/core/frontend/manager/UIManager.cpp
```diff
@@ -46,13 +46,20 @@ namespace YimMenu
 
 	std::shared_ptr<Submenu> UIManager::GetActiveSubmenuImpl()
 	{
-		return m_ActiveSubmenu;
+		if (m_ActiveSubmenu)
```
maybegreat48: What?

src/core/frontend/manager/styles/ModernVertical/ModernVertical.cpp
maybegreat48: You can do that too

src/core/frontend/manager/UIManager.cpp
Deadlineem: I messed that up and caught it after i pushed the changes, already taken care of lol

src/core/frontend/manager/styles/ModernVertical/ModernVertical.cpp
Deadlineem: Done

PR #523, opened 2025-06-21
UI Load window location/size from the configuration file

fangshun2004:
Load from IMGUI.INI Modify the position and size of the window

### DIFF
--- a/src/core/frontend/manager/styles/Classic/Classic.cpp
+++ b/src/core/frontend/manager/styles/Classic/Classic.cpp
@@ -11,8 +11,8 @@ namespace YimMenu
 		float centerY = *YimMenu::Pointers.ScreenResY / 5.0f;
 		ImVec2 windowSize(windowWidth, *YimMenu::Pointers.ScreenResY / 2.5f);
 
-		ImGui::SetNextWindowSize(windowSize, ImGuiCond_Once);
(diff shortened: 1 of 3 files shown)


Deadlineem:
> Load from IMGUI.INI Modify the position and size of the window

https://github.com/YimMenu/YimMenuV2/pull/525 - Nullifies this

I was working on this yesterday just did not push it until i got home from work.  This stores each individual theme settings file in its respective location (if applicable for that theme) designed around using functions like SaveThemeSettings and LoadThemeSettings.  Any time a themes item is moved (Ex: Buttons, Categories, Content) (if the functions are used in said theme) it will update the data instantly in its respective json file.

json files (such as Modular.json) are stored in %APPDATA%\YimMenuV2\Themes\\<Theme Name (Folder)>, for the modular theme file, this would be %APPDATA%\YimMenuV2\Themes\Modular\Modular.json

You can preview this in Modular.cpp or the functions in Themes.cpp to see how they work.  This way imo is way more efficient than using ImGui for it as this not only saves the data but it re-reads it after closing/reopening the game AND it also restores the specific sizes of the theme items as well.  Not only that, but it allows **Fine Tuning** your UI positions/sizes.

Here is the json output from my test of the modular theme
[Modular.json](https://github.com/user-attachments/files/20846154/Modular.json)

PR #531, opened 2025-06-21
fix: update some locals

kurumimeow:
- Update some locals for heists
- Fix misleading globals and locals declaration


kurumimeow:
It probably fixes the crash during heist.

PR #533, opened 2025-06-22
fix: Do not kill the vehicle reward script if we didn't start it

ShinyWasabi:


hjkl123456789:
> _未提供描述。_



> _未提供描述。_

Can you please fix the problem of the Apartment Mission and Cayo Perico getting stuck in the start of the game?

PR #550, opened 2025-06-24
add  fun tab in vehicle submenu

lonelybud:
- add code to raise lowrider wheel individually

Its just a tab to add fun stuff like open doors, toggle neon, get out of car, change radio stations etc. Just want to hear opinion whether to add it? For  now only added lowrider feature.


xesdoog:
Yes please! And if we could port some of the suspension features from VStancer. Mainly camber, wheel offset, and ride height. Also there's an internal game function to break off individual parts (wheels, doors, etc..) I can't do it because I'm dumb.

PR #559, opened 2025-06-25
fix: broken logging under Wine

playday3008:
The problem only is: on Windows it will log a `FATAL` error as it will not find a `wine_get_version` export in `ntdll`, but that's just log:
https://github.com/YimMenu/YimMenuV2/blob/b3d741ca512885d75b83d53f1b5429fc4e012b71/src/core/memory/Module.hpp#L86
https://github.com/YimMenu/YimMenuV2/blob/b3d741ca512885d75b83d53f1b5429fc4e012b71/src/core/memory/Module.hpp#L110
I think this log statement should be removed, or wrap it, so there will be two functions, one `GetExportLoosy` (for example) without log statement, and `GetExport` which will be wrapper around `GetExportLoosy` with log statement. Or create something like `IsExported`, `GetExportImpl`, `GetExport` (here log will be)

### DIFF
--- a/src/core/logger/LogSink.cpp
+++ b/src/core/logger/LogSink.cpp
@@ -4,6 +4,8 @@
 
 #include <format>
 
+#include "core/util/Wine.hpp"
+
 namespace YimMenu
 {
 	LogColor LogSink::GetColor(const eLogLevel level)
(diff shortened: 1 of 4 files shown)


maybegreat48:
What does this aim to fix?

playday3008:
ANSI escape codes aren't working under Wine, it displays them as is.
For example, try to use `FormatConsole` for file, and file output will look basically like console output does under Wine.

playday3008:
Here I pushed the solution for an idea I had about making changes to `Module` class

playday3008:
Yo, @maybegreat48, can you look at this: https://github.com/YimMenu/YimMenuV2/commit/b769d36a1c0460cda677a87a8dcd7eebb035e0ea#r160846324

PR #589, opened 2025-06-29
feat: Allow to spawn ped with me in the vehicle

playday3008:


src/game/frontend/submenus/World/SpawnPed.cpp
```diff
@@ -95,6 +96,13 @@ namespace YimMenu::Submenus
 									handle.SetAsCop();
 								}
 
+								if (spawnInMyVehicle)
+								{
+									auto vehicle = Self::GetVehicle();
+									if (vehicle)
+										handle.SetInVehicle(vehicle, -2);
```
maybegreat48: Never knew about how the game handles -2 before, interesting

playday3008: `-2` means, any available passenger seat

maybegreat48: Yeah, I realized just now after looking at the code, but I wish I'd known about this earlier

playday3008:
Wanted to make a `Bring All Peds Into My Vehicle` (until vehicle is full) button as well, but couldn't make it work, so, meh

playday3008:
Also, maybe it would be better to check whether vehicle have free seats before spawning the ped

kurumimeow:
You can use this to get max number of seats:
`VEHICLE::GET_VEHICLE_MODEL_NUMBER_OF_SEATS(Hash modelHash);`
for every seat check if it's free:
`Vehicle::IsSeatFree(int seat)`

playday3008:
I know how to check it, but the question is "should we check?"

kurumimeow:
Yeah, you should, you only want to put the ped inside the vehicle if any seat is free.

playday3008:
@kurumimeow here you go

kurumimeow:
You can simply do this:
```cpp
if (VEHICLE::ARE_ANY_VEHICLE_SEATS_FREE(vehicle.GetHandle()))
	handle.SetInVehicle(vehicle, -2);
else
	// your notification
```
I thought you wanted to fill all vehicle seats with peds.

playday3008:
> You can simply do this:
> 
> ```c++
> if (VEHICLE::ARE_ANY_VEHICLE_SEATS_FREE(vehicle.GetHandle()))
> 	handle.SetInVehicle(vehicle, -2);
> else
> 	// your notification
> ```

This solution implies that we're spawning ped, and then checking if we can put him in the vehicle, which is bad, because why would we spawn ped to begin with when there's no space for it in the vehicle

> I thought you wanted to fill all vehicle seats with peds.

It's still doable, I didn't remove that functionality

PR #606, opened 2025-07-04
fix: fix BattlEyeServerProcessPlayerJoin

Mr-X-GTA:


maybegreat48:
This is definitely breaking next update but would do for now. Thanks for this; if we don't hook this correctly people will get shadowbanned much more often

maheswar192:
@maybegreat48 @Mr-X-GTA Is it safe for now ??? and what's new on this update

JJGUY123456:
Is the new GTA update related to this? Cause if it is, it was really fast.

PR #609, opened 2025-07-04
[Added] Loud  Subwoffer for Tuning Meet

CSY0N:
**this option has this awesome feature**

- _Enables loud radio on vehicles, Players can hear it, perfect for LSCM or Tuning Meet_


SourceModzZ:
I tested it with my friend, but unfortunately he can't hear the car radio from my car.

xesdoog:
> I tested it with my friend, but unfortunately he can't hear the car radio from my car.

Because this only increases the volume slightly (lowriders with the trunk subwoofer upgrade from benny's are much louder than this). If you want something really loud, static emitters are the best approach.

EDIT: Also, this is bad. He's looping over all vehicles and calling these 3 natives, every frame:

```
AUDIO::SET_VEHICLE_RADIO_ENABLED
AUDIO::SET_VEHICLE_RADIO_LOUD
VEHICLE::SET_VEHICLE_ENGINE_ON
```

Deadlineem:
> > I tested it with my friend, but unfortunately he can't hear the car radio from my car.
> 
> Because this only increases the volume slightly (lowriders with the trunk subwoofer upgrade from benny's are much louder than this). If you want something really loud, static emitters are the best approach.
> 
> EDIT: Also, this is bad. He's looping over all vehicles and calling these 3 natives, every frame:
> 
> ```
> AUDIO::SET_VEHICLE_RADIO_ENABLED
> AUDIO::SET_VEHICLE_RADIO_LOUD
> VEHICLE::SET_VEHICLE_ENGINE_ON
> ```

I tested this out, it does seem to make the radio for all nearby vehicles have more bass to them (walk by a group of npc cars with this enabled and the music volume at 100) but I am unsure if this only runs locally or not.

xesdoog:
> I tested this out, it does seem to make the radio for all nearby vehicles have more bass to them (walk by a group of npc cars with this enabled and the music volume at 100).

You've had this option with a very similar name (except the unnecessary loop) in Extras Addon for a long time. Yes, it does make a slight difference but there's absolutely no need to call it on tick or on every nearby vehicle. You could simply call it once on your own vehicle.

> but I am unsure if this only runs locally or not

I tested this over a year ago with TokyoDrift and iirc it was synced. I faintly remember testing both this and the DJ animations that used to spawn an invisible Party Bus. Actually it was you who helped me test the animations (the party bus was synced too, naturally).

Deadlineem:
> > I tested this out, it does seem to make the radio for all nearby vehicles have more bass to them (walk by a group of npc cars with this enabled and the music volume at 100).
> 
> You've had this option with a very similar name (except the unnecessary loop) in Extras Addon for a long time. Yes, it does make a slight difference but there's absolutely no need to call it on tick or on every nearby vehicle. You could simply call it once on your own vehicle.
> 
> > but I am unsure if this only runs locally or not
> 
> I tested this over a year ago with TokyoDrift and iirc it was synced. I faintly remember testing both this and the DJ animations that used to spawn an invisible Party Bus. Actually it was you who helped me test the animations (the party bus was synced too, naturally).

Interesting.  Honestly I have been working on the UI stuff more than anything lately, but I have referenced a bunch of our work in extras addon to others over time but i forgot about the radio options (i never used the options much mostly building/testing), its mad helpful for features when I decide to work on adding more in enhanced

src/game/features/vehicle/BigLoudSubwoofer.cpp
```diff
@@ -0,0 +1,29 @@
...
+namespace YimMenu::Features
+{
+	class AllVehsLoudRadio : public LoopedCommand
+	{
+		using LoopedCommand::LoopedCommand;
+
+		virtual void OnTick() override
+		{
+			for (auto veh : Pools::GetVehicles())
+			{
```
maybegreat48, line 15: Why are we enabling it for all vehicles?

xesdoog: Shouldn't be OnTick either. This only works **accidentally** because he's looping it. Setting the engine on at the end makes no sense, it should be at the top with a short yield but still, there's no need for all of that. All he needs to do is call `SET_VEHICLE_RADIO_LOUD` on the player's vehicle only once on enable. If he wants it to persist when the player changes vehicles then he will have to manually track it by adding a member `Vehicle::m_HasLoudRadio`.

CSY0N:
@maybegreat48 any issues? of the Code when yeah tell me

AkumanoRev:
excuse me, can you add replay cutscene from first dlc until money front? please

PR #623, opened 2025-07-10
fix: track and cleanup patched vtables to prevent memory leak on unload

SourceModzZ:


maybegreat48:
Can the menu even unload properly?

SourceModzZ:
> Can the menu even unload properly?

Correctly recognized: currently the menu cannot be unloaded, so the cleanup is currently not usable - but it makes the code future-proof and shows correct resource management.

PR #418, opened 2025-06-12
(Improvement/Feat) UI Improvements, Casino Bypass Cooldown, Icon Support, Remove Cyrillic, Etc.

Deadlineem:
For slot machines and tables, bypasses the cooldown/ban so you can play as much as you want to.

Screenshot_111

Update:
Screenshot_112

PR #657, opened 2025-07-30
feat(Recovery): add casinoWheel

Cleanwn:


CSY0N:
looks risky | and Risky things will not be added in YimV2 but you can use it for yourself privat and have you tested it bevor you suggest it here?

SourceModzZ:
Maybegreat certainly won't add it, as this method is very risky. And maybegreat doesn't want risky things in YimV2.

eXhumer:
Has this been tested?

https://github.com/YimMenu/YimMenuV2/pull/418#issuecomment-2973141292

Based on comment above, bypassing lucky wheel cooldown causes transaction errors.

PR #721, opened 2025-09-21
Add the comparison features for packed and generic statistics under the statistical editor.

fangshun2004:
1.Increased the editing and importing of date, pos, and userid types.
2.Added the feature for comparing packed and supported statistical data.

### DIFF
--- a/src/game/frontend/submenus/Recovery/StatEditor.cpp
+++ b/src/game/frontend/submenus/Recovery/StatEditor.cpp
@@ -11,6 +11,10 @@
 #include <utility>
 #include <vector>
 
+
+using stats_vec = std::vector<std::pair<std::string, std::string>>;
+using packed_vec = std::vector<std::pair<uint32_t, uint8_t>>;
+
 namespace YimMenu::Submenus
 {
 	struct StatInfo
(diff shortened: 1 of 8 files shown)


fangshun2004:
The comparison results of the packed statistics are in cmp_packed_date.txt of YimMenuV2 directory
The stats comparison requires the files containing statistical data names to be named stats.txt and placed in the YimMenuV2 directory. The comparison results will be found in the YimMenuV2's cmp_stats_date.txt file.
Use "//" or a blank line to skip while ignoring the string after ","
stats file containing all statistical data
[stats.txt](https://github.com/user-attachments/files/22452447/stats.txt)

SourceModzZ:
Very good work. But I think this repo is dead.

src/game/frontend/submenus/Recovery/StatEditor.cpp
```diff
@@ -182,27 +232,30 @@ namespace YimMenu::Submenus
 				return tolower(c);
 			});
 
-			if (as_string == "true" || as_string == "1")
+			if (as_string != "false" || as_string != "0")
```
xiaoxiao921: If you invert the condition like that then you also need to change || to &&

PR #774, opened 2025-12-17
fix: Fix Manipulate Rig Slot Machines for b1013.17

M3351AN:


src/game/features/recovery/Casino.cpp
```diff
@@ -15,8 +15,9 @@ namespace YimMenu::Features
 	{
 		using LoopedCommand::LoopedCommand;
 
-		int slots_random_results_table = 1374;
+		int slots_random_results_table = 1350;
```
ShinyWasabi, line 18: I couldn't find any usage of 1350 in the script in either 1013.20 or 1013.17. What is it?

M3351AN: In simple terms, this is a two-dimensional array that "maybe" has been "optimized during compilation". When passed as param, `uLocal_1374` is passed, but for patching/reversing, the actual start of the data struct is not at `uLocal_1374`.

In simple terms, we found that references to `uLocal_1374` often (should be "always" if my thought is right) take the form of `uLocal_1374.f_1[0 /*65*/][uLocal_2117[0]]`, `uLocal_1374.f_278[1]`, etc. However, in the decompiled casino-slots.ysc file, we don't know what f_1 or f_278 actually represent (at least within this single file)

But we decompiled the files from the old version of "which old code ever working". In the old version (1.71), the same functions referenced `uLocal_1372.f_1[...][...]`. We assumed that the entire data structure was shifted backwards by 2 bytes, so 1348 -> 1350 

And according to testing, this worked.
0e3f89644163967fbc8c8ec2b5f2b355

Issue #601, opened 2025-07-02
[Bug]: Personal Vehicles not showing


Undefxned:
### Describe the bug

In Vehicle > Spawn > Personal Vehicle some vehicles aren't showing.
I noticed that sometimes they will show, sometimes not. And at the next restart it might happen again with other Personal Vehicles that were showing just fine before, it appears to be random.
EDIT: If you own two or more vehicles of the same model, only one will be showed in the mod menu selection. This is a persistent bug even after changing session/restarting the game.

### Store

Rockstar

### Anticheat bypass mode

Without FSL

### Build

- [x] I've confirmed the issue exists on the latest version of YimMenuV2

### Duplicate

- [x] I confirm that issue isn't a duplicate.

Poopybuttho1e:
i noticed this mainly with the vinewood garage, it only shows the first 10 cars for the first floor and then it bugs out and shows only 3 of the cars from the first floor except in the menu the second floor is selected

<img width="1586" height="1036" alt="Image" src="https://github.com/user-attachments/assets/21e492f8-4203-4483-ae45-75892ce2799e" />


You can see in the second pic the license plate for the kanjo sj matches even though basement level 2 is selected and im on the first floor. My actual level 2 garage has a whole bunch of different cars which are being shown in level 3 and isnt showing the full 17 cars either
<img width="1426" height="1034" alt="Image" src="https://github.com/user-attachments/assets/0527b824-5f92-4faf-80fa-6f9063658049" />

Undefxned:
> i noticed this mainly with the vinewood garage, it only shows the first 10 cars for the first floor and then it bugs out and shows only 3 of the cars from the first floor except in the menu the second floor is selected
> 
[...]

PR #793, opened 2025-12-20
Fix the garage missing some vehicles

fangshun2004:
Only one vehicle with the same model and license plate will be displayed in the fix menu
#601


Undefxned:
Thanks for your great work.

L7NEG:
gwkiu = great work keep it up

PR #819, opened 2025-12-25
Implement ScriptLocal class for Lua scripting

Xiseous:
"ScriptLocal" provides access to **local variables** within running scripts. 

Unlike "ScriptGlobal" which accesses global variables shared across all scripts, "ScriptLocal" accesses variables that are specific to a particular script instance's execution context.

### DIFF
--- a/src/game/scripting/libraries/ScriptLocal.cpp
+++ b/src/game/scripting/libraries/ScriptLocal.cpp
@@ -1,0 +1,117 @@
+#include "core/scripting/LuaLibrary.hpp"
+#include "core/scripting/LuaScript.hpp"
+#include "core/scripting/LuaUtils.hpp"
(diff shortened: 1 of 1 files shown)


nooksmr:
what differ from lua "local" variables?

example with bizteroids
{B917C3CD-B705-4919-A6BD-51BFD619FB7C}

Xiseous:
> what differ from lua "local" variables?
> 
> example with bizteroids {B917C3CD-B705-4919-A6BD-51BFD619FB7C}

Use ChatGPT and ask it “what does set local do in a LUA script” because these are completely different   things, also read the original pull comment on what the function does.

ScriptLocal would let you, for example be able to edit on-going game loops and create instant finish LUA scripts, to for example, instant finish a delivery mission.

L7NEG:
LGTM

PR #843, opened 2026-01-03
fix(anticheat): fix anticheat bypass

Mr-X-GTA:


hjkl123456789:
> _未提供描述。_

Thank you so much for fixing the YimMenuV2 anti-cheat bypass vulnerability. Now I can enable anti-cheat and play with all players in public lobbies without being kicked out. When doing heist missions, inviting others doesn't get stuck during the mission startup. You did a fantastic job!

PR #899, opened 2026-02-24
Fix(core): Unload crash fixes and unload button

Tessio:
- main.cpp: fixed destruction order (ScriptMgr→NativeHooks→FiberPool→Hooking)
 - AnticheatBypass.cpp: prevent duplicate hooks with HooksInitialized
 - Hooking.cpp: improved cleanup process
 - DetourHook.hpp: use MH_RemoveHook in destructor
 - NativeHooks.cpp: clear registered programs and hooks
 - ScriptMgr.cpp: add 100ms delay for clean script shutdown
 - Settings.cpp: add unload button in Hotkeys category


xesdoog:
I think you should add this fix to [TemplateHellBase](https://github.com/Yimura/TemplateHellBase) if you don't mind.

Tessio:
> I think you should add this fix to [TemplateHellBase](https://github.com/Yimura/TemplateHellBase) if you don't mind.

Unless this is sarcasm that I have failed to detect, isn't that the main “forgotten” base of this project?

xesdoog:
> Unless this is sarcasm that I have failed to detect, isn't that the main “forgotten” base of this project?

It's a template repository that this menu and HorseMenu were generated from and both have the same unload crash issue. The template is actually a very solid base and I've used it for two different projects completely unrelated to GTA and the RAGE engine. Both also had the same unload issue which means the problem is in the base itself, hence the request to add the fix to the template.

Tessio:
> > Unless this is sarcasm that I have failed to detect, isn't that the main “forgotten” base of this project?
> 
> It's a template repository that this menu and HorseMenu were generated from and both have the same unload crash issue. The template is actually a very solid base and I've used it for two different projects completely unrelated to GTA and the RAGE engine. Both also had the same unload issue which means the problem is in the base itself, hence the request to add the fix to the template.

As the mechanic would say, It's comin' at ya!

0xEmiil:
Waitt, why we are adding ms on cleanup on x64, i mean for x32 i guess its reliable, but on x64 hummm......

xiaoxiao921:
It's reliable on neither, the proper way is to sync it through shared variables or signaling mechanisms, but it's better than nothing I guess

0xEmiil:
Told ya, ​🤷‍♂️​

PR #933, opened 2026-05-13
chore: Update data hash.

tupoy-ya:
Currently the `spoofdatahash` feature will always put you into empty sessions because it uses the hash from the previous update.


kudrnina2013:
Please help me urgently, my game crashes when I inject cheats

PR #949, opened 2026-06-17
feat(lua): UI Support and Other Implementations

tuyilmaz:
Lua API:
- New libraries/improvements: menu, commandmgr, memory, FileMgr, pointer, ImGui, Vector3, ScriptGlobal/Local/Pointer/Data/Function, Player/players, entities, event, scr_patch
- Per-script UI ownership with full teardown on script unload

Natives:
- Support pointer arguments
- Emit a LuaLS natives stub (docs/natives.lua) from GenerateNatives.py

Docs:
- Add docs/yimmenu_v2.lua(AI Generated) and docs/natives.lua


src/game/scripting/libraries/ScriptData.cpp
```diff
@@ -0,0 +1,101 @@
+#include "core/memory/Pattern.hpp"
```
ShinyWasabi: I don't think we need to expose this to Lua. It purely exists for script patches. We already have `ScriptPointer`, which you've exposed.

ShinyWasabi (approved):

PR #972, opened 2026-07-14
fix(pointers): update patterns after last update

m4nst3in:
fixed some patterns that broke after the last update

SendNetworkDamage - jnbe is near now, still sub 0x51
PrepareMetricForSending - F9 to F1, sub 0x26 to 0x4B
SendEventAck - new sig, EventAck still Sub(4).Rip, SendEventAck Add(0x15).Add(1).Rip
BattlEyeStatusUpdate - patch entry at +0x10C (xor eax,eax; ret)
JoinSessionByInfo - entry prolog
HandleJoinRequestIgnorePoolPatch - new follow-up after the jz


m4nst3in:
closing, reopening with the fixed patterns

PR #979, opened 2026-07-14
Fix Casino Variables

thororen1234:
just fixes the casino variables for new update


thororen1234:
https://github.com/YimMenu/YimMenuV2/pull/988

PR #985, opened 2026-07-15
fix: 12 high-risk bugs across hooking, renderer, memory, scripting

prathamtagad:
## Summary

Identified and fixed 12 high-risk bugs across the codebase that cause crashes, 
use-after-free, memory corruption, and data races.

### Critical (Crash/Corruption)
| File | Bug | Fix |
|------|-----|-----|
| `Renderer.hpp` | REL macro calls Release() on null | Check null before Release() |
| `Renderer.cpp` | FrameContext copied by value, fence never consumed | Changed to reference |
| `main.cpp` | `getenv("appdata")` can return nullptr → UB | Added null guard |
| `DetourHook.hpp` | `m_Enabled` never set → hooks never enable/disable properly | Set flag after operations |
| `IATHook.hpp` | IAT write without VirtualProtect → crash on RO pages | Added VirtualProtect |
| `Module.cpp` | `SizeOfCode` only covers .text → patterns in other sections missed | Use `SizeOfImage` |
| `ExceptionHandler.cpp` | Zeroing RSP → guaranteed secondary crash | Return CONTINUE_SEARCH |
| `LuaScript.cpp` | `return` instead of `continue` → resources never cleaned up | Changed to `continue` |

### High (Performance/Safety)  
| File | Bug | Fix |
|------|-----|-----|
| `PatternScanner.cpp` | No break on mismatch → 2-5x slower scanning | Added `break` |
| `DetourHook.hpp` | `throw` after error → dead code, unsafe in DLL context | LOGF + return |
| `CallHook.cpp` | Fixed VirtualAlloc address, no null check | Fallback + error log |
| `Notifications.cpp` | TOCTOU race on duplicate notification check | Lock before check |

## Testing
- [x] Compiles successfully
- [x] All changes are minimal and surgical - no refactoring

### DIFF
--- a/src/core/hooking/CallHook.cpp
+++ b/src/core/hooking/CallHook.cpp
@@ -5,6 +5,13 @@ namespace YimMenu
 	CallHookMemory::CallHookMemory()
 	{
 		m_Memory = VirtualAlloc((void*)((uintptr_t)GetModuleHandle(0) + 0x40000000), 1024, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE);
+		if (!m_Memory)
+			m_Memory = VirtualAlloc(nullptr, 1024, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE);
+		if (!m_Memory)
+		{
+			LOGF(FATAL, "CallHookMemory: Failed to allocate jump sequence memory");
+			throw std::runtime_error("CallHookMemory: Failed to allocate jump sequence memory");
+		}
 		m_Offset = 0;
 	}
 
--- a/src/core/hooking/IATHook.hpp
+++ b/src/core/hooking/IATHook.hpp
@@ -42,6 +42,7 @@ namespace YimMenu
 	inline bool IATHook<T>::Enable()
 	{
 		*m_HookLocation = m_HookFunc;
+		m_Enabled = true;
 		return true;
 	}
 
@@ -49,6 +50,7 @@ namespace YimMenu
 	inline bool IATHook<T>::Disable()
 	{
 		*m_HookLocation = m_OriginalFunc;
+		m_Enabled = false;
 		return true;
 	}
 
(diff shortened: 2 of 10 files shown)


src/core/hooking/IATHook.hpp
```diff
@@ -41,14 +41,22 @@ namespace YimMenu
 	template<typename T>
 	inline bool IATHook<T>::Enable()
 	{
+		DWORD oldProtect, temp;
+		VirtualProtect(m_HookLocation, sizeof(void*), PAGE_READWRITE, &oldProtect);
```
Mr-X-GTA: This is not required for GTA 5 because the text section is already RWX

src/core/logger/ExceptionHandler.cpp
```diff
@@ -100,7 +100,9 @@ namespace YimMenu
 					case 1: exception_info->ContextRecord->Rcx = 0; break;
 					case 2: exception_info->ContextRecord->Rdx = 0; break;
 					case 3: exception_info->ContextRecord->Rbx = 0; break;
-					case 4: exception_info->ContextRecord->Rsp = 0; break;
+					case 4:
+					exception_info->ContextRecord->Rip -= opcode.len;
+					return EXCEPTION_CONTINUE_SEARCH; // zeroing RSP would cause an immediate secondary crash
```
Mr-X-GTA: Revert this

prathamtagad: Done 👍

jotkauser:
AI slop PR description and the code is probably AI too

src/core/hooking/CallHook.cpp
```diff
@@ -5,6 +5,13 @@ namespace YimMenu
 	CallHookMemory::CallHookMemory()
 	{
 		m_Memory = VirtualAlloc((void*)((uintptr_t)GetModuleHandle(0) + 0x40000000), 1024, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE);
+		if (!m_Memory)
+			m_Memory = VirtualAlloc(nullptr, 1024, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE_READWRITE);
```
Mr-X-GTA, line 9: The address cannot be more than INT32_MAX away from the instruction after the call instruction

PR #1037, opened 2026-07-23
feat: Hide Hud Toggle + Mute HUD UI toggle

mrbouquet69:
add new buttons to settings menu to mute hud ui text messages invites etc and add a button to show and hide hud like minimap


mrbouquet69:
gotta redo i edited wrong settings file

PR #1047, opened 2026-07-24
feat: harden network parsing and add runtime diagnostics

wyattfr:
## Summary

- bounds-check bit-buffer reads, writes, seeks, strings, floats, and network-controlled payload sizes
- validate scripted events, packed events, packed reliables, and BattlEye commands before accessing their data
- add a privacy-safe Network → Diagnostics panel with compatibility state, FSL capabilities, protection totals, last-seen timestamps, reset, and clipboard export
- recover corrupt settings from an atomic backup and make serializer dirty flags thread-safe
- replace the raw pattern cache with a bounded, versioned, atomically replaced format
- fix Lua resource lifecycle, hotkey edge handling, exception logging synchronization, main-loop CPU use, and UI width narrowing warnings
- add bit-buffer and concurrent telemetry regression tests and run them in Windows CI

## Why

Several network-controlled lengths were trusted before packet and destination bounds were checked. The bit-buffer implementation also contained incorrect 64-bit and signed serialization logic. Persistence paths could leave corrupt settings or cache data after interrupted writes, while there was no user-facing way to inspect compatibility or protection activity.

## Impact

Malformed packets are rejected without overrunning buffers or hanging parsers. Users gain an aggregate diagnostics view that stores no player identifiers and can copy a support report directly to the clipboard. Settings can recover from their last known-good backup, the pattern cache validates its format, and runtime state updates are safer across threads.

## Validation

- configured with CMake using MSVC 19.44
- built `YimMenuV2.dll` successfully
- `dat_bit_buffer_tests`: passed
- `protection_telemetry_tests`: passed, including concurrent increments
- 2/2 CTest tests passed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Build and run unit tests
         run: |
-          cmake --build ./build --config RelWithDebInfo --target dat_bit_buffer_tests --
+          cmake --build ./build --config RelWithDebInfo --target dat_bit_buffer_tests protection_telemetry_tests --
           ctest --test-dir ./build --build-config RelWithDebInfo --output-on-failure
 
       - name: Check if DLL got built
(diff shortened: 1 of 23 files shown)


ShinyWasabi:
enough with the AI slop

wyattfr:
You can give me some ai slop 😋

Issue #1065, opened 2026-08-01
[Request]: teleport forward


Pluzas:
### Problem

this could be useful doing casino, i can jus tp forward before drilling the vault and starting the timer, so i can grab everything and get 100% of cut, it used to be a feature in legacy version and i feel like the teleport section is poor overall, missing many cool tp features 

### Solution

­

### Reason

­

### Duplicate

- [x] I confirm that issue isn't a duplicate.

### !!!READ THIS!!! Recovery features

- [x] I confirm that I'm not requesting a recovery feature

madjidcorso:
Use NoClip to move forward

L7NEG:
> Use NoClip to move forward

Exactly https://github.com/YimMenu/YimMenuV2/pull/1067#issuecomment-5182046644

PR #1067, opened 2026-08-02
feat: add ImGui render callback functions and directional teleport

tuyilmaz:
LUA:
- Added add_imgui and add_always_draw_imgui.
- Added get_pointer to ScriptLocal and ScriptGlobal classes.

General:
- Added directional tp commands  #1065


L7NEG:
LGTM but can't we just use noclip atp ?

tuyilmaz:
> LGTM but can't we just use noclip atp ?

Yes, the main purpose of this pr is the lua api additions. That's just an extra feature, wouldn't hurt. We had both noclip and directional tp in V1.

L7NEG:
> > LGTM but can't we just use noclip atp ?
> 
> Yes, the main purpose of this pr is the lua api additions. That's just an extra feature, wouldn't hurt. We had both noclip and directional tp in V1.

ok makes sense now :D.

PR #1069, opened 2026-08-03
fix （‎StatEditor）Fixed the issue with writing 64-bit statistics in the Statistics Editor

fangshun2004:
## Current behavior:
1. After modifying the int64 data type, restarting the game will restore the data. 
2. Modify the UINT64 data type. If the data exceeds the maximum value of int, the modified data will differ from the read value, and when writing again, it will overflow to the maximum value of UINT64

## Changes Proposed:
1.Fixed data overflow after writing data to uint64.
2.Fixed the issue where int64 data would recover after writing and saving and reopening the game.
3.Add packed int/bool stats for all writes and reads.
4.Add DATE POS USERID  stats writes and reads.

### DIFF
--- a/src/game/frontend/submenus/Recovery/StatEditor.cpp
+++ b/src/game/frontend/submenus/Recovery/StatEditor.cpp
@@ -39,7 +39,7 @@ namespace YimMenu::Submenus
 		int m_AsInt;
 		bool m_AsBool;
 		std::uint64_t m_AsU64;
-		char m_AsString[64];
+		char m_AsString[21];
 		Date m_Date;
 	};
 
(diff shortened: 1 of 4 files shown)


fangshun2004:
//bit0-bit49 true ，50 jump points complete the position
MPX_USJS_COMPLETED_MASK=1125899906842623
Story mode:
USJS_COMPLETED_MASK=1125899906842623
test

src/game/frontend/submenus/Recovery/StatEditor.cpp
```diff
@@ -208,20 +225,24 @@ namespace YimMenu::Submenus
...
-			STATS::STAT_SET_MASKED_INT(hash, (std::uint32_t)uint64_, 0, 32, true);
-			STATS::STAT_SET_MASKED_INT(hash, (std::uint32_t)(uint64_ >> 32), 32, 32, true);
+			data->SetUInt64(uint64_ - 1);
+			STATS::STAT_INCREMENT(hash, static_cast<float>(1));
 			return;
 		}
 		case sStatData::Type::STRING:
 			STATS::STAT_SET_STRING(hash, value.data(), true);
 			return;
+		case sStatData::Type::PACKED:
```
ShinyWasabi, line 285: You need to put this case in braces (`initialization of 'uint64_' is skipped by 'default' label`).

ShinyWasabi (changes requested):

PR #1084, opened 2026-08-15
Rewrite Renderer

ShinyWasabi:
Fixes the AMD FSR 3 Frame Generation crash


MxHassan195:
that's really nice

fangshun2004:
Do programs uninstalling part of the main.cpp not need to use D3D12Hook::D estroyImpl？

PR #1109, opened 2026-08-22
Vehicle Menu & Management Improvements

rkwapisz:
I like the ability to save vehicles to json with YimMenuV2, but the current implementation is not particularly user-friendly and needs some polishing. Most of these changes are subjective, but I'm confident that they improve the user experience.

Here is a screenshot of how the reworked menu looks:
image

Summary of changes:

- Rebuilt Saved Vehicles page for a cleaner and more organized look

- More verbose text feedback throughout the vehicle save workflow

- Removed cumbersome "do you want to spawn" confirmation dialog when selecting a vehicle from the list; now it's just a single click is required to spawn the vehicle. Spawning a vehicle is a low-risk low-impact event, so we don't need a confirmation dialog.

- Max Vehicle option in the Vehicle Editor is more comprehensive in applying its modifications (now includes tinted windows, xenon + neon lights, and a custom horn) and also repairs the vehicle

- Reworked "burstible tires" to "bulletproof tires" to better match in-game customization nomenclature and identify this feature as an upgrade rather than a default

- Vehicle Editor's "Extras" will no longer display a separator and blank space if there are no extras available to the vehicle

- Fixed the Vehicle Editor's color picker inconsistency due to RBG and HSV conflicts (only one of these flags should be pushed into ImGui's color picker)

### DIFF
--- a/src/game/frontend/submenus/Vehicle/SavedVehicles.cpp
+++ b/src/game/frontend/submenus/Vehicle/SavedVehicles.cpp
@@ -13,61 +13,56 @@ namespace YimMenu::Submenus
 {
 	static BoolCommand spawnInsideSavedVehicle{"spawninsidesavedveh", "Spawn Inside", "Spawn inside the vehicle."};
 
+	namespace
+	{
+		// Get rid of the ".json" file extension so keep the saved car list a little cleaner
+		std::string GetDisplayName(const std::string& fileName)
+		{
+			constexpr std::string_view extension = ".json";
+			if (fileName.size() > extension.size() && std::string_view(fileName).ends_with(extension))
+				return fileName.substr(0, fileName.size() - extension.size());
+
+			return fileName;
+		}
+
+		std::string ToLower(std::string value)
+		{
+			std::transform(value.begin(), value.end(), value.begin(), [](unsigned char c) {
+				return static_cast<char>(std::tolower(c));
+			});
+			return value;
+		}
+	}
+
 	std::shared_ptr<Category> BuildSavedVehiclesMenu()
 	{
 		static std::string folder{}, file{};
 		static std::vector<std::string> folders{}, files{};
-		static char vehicle_file_name_input[64]{};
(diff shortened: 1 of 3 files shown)


lmagineNothing:
This looks very nice! :)
I think this could also work well, just an idea:
quickchange

rkwapisz:
Let me see if I understand what you're proposing.

In this PR, there is a Folder Name text field if the folder selected is Root. If there is a different subfolder selected, then the Folder Name field disappears and the vehicle will be saved in that subfolder.

image

Are you asking for the ability to create additional nested subfolders?

CSY0N:
is useless for real

lmagineNothing:
> Let me see if I understand what you're proposing...

Sorry, my last comment was too vague, let me clarify.
In your first screenshot, "File Name" is above the "Name this vehicle" box. I think that moving File Name to the right side of the box would make that part look more _aligned_.

You can ignore the second screenshot I added, it was just an example of how it would look when the root folder is selected.

lonelybud:
Code wise looks LGTM, Haven't tested the UI. Other reviewers might point out extra code in `Vehicle::Upgrade` function. That 4 line logic to check if that tab was opened is clever but maybe something like `onTabClick` general function should be there for every tab. Overall fine.

xesdoog:
> Sorry, my last comment was too vague

The author **deliberately** made the label separate from the input text widget and above it. In ImGui, widget labels are always placed on the right side and when a dev chooses not to show the label or show it somewhere else, they use an anonymous label *(a string that starts with two pound symbols, ex: "##filename")* in the label parameter and place their actual label wherever using an `ImGui::Text`. You're asking them to revert to the default ImGui behavior instead to align with the "Folder Name" input text which makes sense, both widgets should have the same layout.

MrEonn11:
What about a list of spawned cars area that we can select the cars we've spawned and delete them or teleport in them etc? Just an idea came into my mind.

PR #1110, opened 2026-08-23
fix: drop ped-deletion PackedReliables instead of mis-patching object…

SourceModzZ:
Writing a 16-bit ID into a 13-bit field corrupted the packet and crashed in the original handler after a blocked player deletion.


Mr-X-GTA:
We set the object ID to an ID that doesn't exist to block the remove, because blocking it directly would require us to send an ack. The next 5 bits after the object ID are the ownership token, which isn't even read if the object doesn't exist.

SourceModzZ:
> We set the object ID to an ID that doesn't exist to block the remove, because blocking it directly would require us to send an ack. The next 5 bits after the object ID are the ownership token, which isn't even read if the object doesn't exist.

So i have to close pull request ?

